PR #1, opened 2022-11-02
Adds template from cookiecutter pytest plugin

AlejandroFernandezLuces:
### Overview

Adds a skeleton for the project, generated by the [pytest template tool](https://github.com/pytest-dev/cookiecutter-pytest-plugin/blob/main/README.md). I believe it is better to have this skeleton merged to main first, so next PR modifications are clearly differentiated from the template.

Everything from the template should be fine, but you might want to review if there is anything that needs to be changed.


appveyor.yml
```diff
@@ -0,0 +1,40 @@
...
+# https://www.appveyor.com/docs/build-environment/#python
+
+environment:
+  matrix:
+    - PYTHON: "C:\\Python35"
+      TOX_ENV: "py35"
+
+    - PYTHON: "C:\\Python36"
+      TOX_ENV: "py36"
+
```
MaxJPRey, line 11: @AlejandroFernandezLuces Most of the project (PyVista, Pyansys, Numpy...) are not supporting version prior to 3.7.
I think here we can get rid of those versions (3.5 and 3.6).
```suggestion
```

MaxJPRey, line 11: @AlejandroFernandezLuces Most of the project (PyVista, Pyansys, Numpy...) are not supporting version prior to 3.7.
I think here we can get rid of those versions (3.5 and 3.6).
```suggestion
```

tox.ini
```diff
@@ -0,0 +1,12 @@
+# For more information about tox, see https://tox.readthedocs.io/en/latest/
+[tox]
+envlist = py35,py36,py37,py38,pypy3,flake8
```
MaxJPRey, line 3:
```suggestion
envlist = py37,py38,pypy3,flake8
```

setup.py
```diff
@@ -0,0 +1,50 @@
...
+
+def read(fname):
+    file_path = os.path.join(os.path.dirname(__file__), fname)
+    return codecs.open(file_path, encoding='utf-8').read()
+
+
+setup(
+    name='pytest-pyvista',
+    version='0.1.0',
+    author='PyAnsys',
```
MaxJPRey, line 17: Should be PyVista here
```suggestion
    author='PyAnsys',
```

setup.py
```diff
@@ -0,0 +1,50 @@
...
+    install_requires=['pytest>=3.5.0'],
+    classifiers=[
+        'Development Status :: 4 - Beta',
+        'Framework :: Pytest',
+        'Intended Audience :: Developers',
+        'Topic :: Software Development :: Testing',
+        'Programming Language :: Python',
+        'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.5',
+        'Programming Language :: Python :: 3.6',
```
MaxJPRey, line 36:
```suggestion
```

setup.py
AlejandroFernandezLuces: Fixed, I couldn't apply your suggestion because you wrote PyAnsys again 😅

MaxJPRey: Fixed. Thanks. I did not apply my suggestion originally.

AlejandroFernandezLuces: I fixed it in a previous commit, it is solved now 😄

MaxJPRey (approved):
LGTM.

PR #3, opened 2022-11-07
Add verify_image_cache fixture from PyVista

AlejandroFernandezLuces:
## Overview
Adds the code from PyVista related to image cache testing to the plugin. Here I expose what will be the main fixture of the plugin, `verify_image_cache`, which will check if the cached image and the generated figures are similar enough. 

I also added the pre-commit files from PyVista, to have the same code format and rules. 

## Details
- Removed lists with the name of the tests, now each test will declare if it can run on windows/MACos and if it is a high variance test

## References
https://github.com/pyvista/pyvista/blob/main/tests/plotting/test_plotting.py


MaxJPRey:
@AlejandroFernandezLuces , I would suggest to add a more descriptive title for the PR.

README.rst
```diff
@@ -43,8 +46,62 @@ You can install "pytest-pyvista" via `pip`_ from `PyPI`_::
...
+
+- high_variance_tests:  If necessary, the threshold for determining if a test will pass or not is 
+incremented to another predetermined threshold. This is currently done due to the use of an unstable 
+version of VTK, in stable versions this shouldn't be necessary.
+
+- windows_skip_image_cache: For test where the plotting in Windows is different from MacOS/Linux.
+
+- macos_skip_image_cache: For test where the plotting in MacOS is different from Windows/Linux.
+
+- skip: If you have a test that plots a figure, but you don't want to compare it's output against the cache,
```
MaxJPRey, line 101:
```suggestion
- skip: If you have a test that plots a figure, but you don't want to compare its output against the cache,
```

README.rst
```diff
@@ -43,8 +46,62 @@ You can install "pytest-pyvista" via `pip`_ from `PyPI`_::
 
 Usage
 -----
+Once installed, you only need to have a `pl.show()` in your test, the plugin will automatically manage the cache generation if it does not exists,
```
MaxJPRey, line 49:
```suggestion
Once installed, you only need to use the command `pl.show()` in your test and the plugin will automatically manage the cache generation if it does not exist,
```

pytest_pyvista.py
```diff
@@ -1,21 +1,168 @@
...
+        action="store_true",
+        help="Enables failure if image cache does not exist.",
+    )
+    parser.addini("HELLO", "Dummy pytest.ini setting")
+
+
+class VerifyImageCache:
+    """Control image caching for testing.
+
+    Image cache files are names according to ``test_name``.
```
MaxJPRey, line 36:
```suggestion
    Image cache files are named according to ``test_name``.
```

pytest_pyvista.py
```diff
@@ -1,21 +1,168 @@
...
+            warnings.warn(
+                f"{test_name} Exceeded image regression warning of "
+                f"{allowed_warning} with an image error of "
+                f"{error}"
+            )
+
+
+@pytest.fixture(autouse=True)
+def verify_image_cache(request, pytestconfig):
+    """Checks cached images against test images for PyVista"""
```
MaxJPRey, line 157:
```suggestion
    """Checks cached images against test images for PyVista."""
```

pytest_pyvista.py
```diff
@@ -1,21 +1,168 @@
...
+    Image cache files are names according to ``test_name``.
+    Multiple calls to an instance of this class will append
+    `_X` to the name after the first one.  That is, files
+    ``{test_name}``, ``{test_name}_1``, and ``{test_name}_2``
+    will be saved if called 3 times.
+
+    Parameters
+    ----------
+    test_name : str
+        Name of test to save.  Sets name of image cache file.
```
MaxJPRey, line 45:
```suggestion
        Name of test to save.  It is used to define the name of image cache file.
```

pytest_pyvista.py
```diff
@@ -1,21 +1,168 @@
...
+
+        # some tests fail when on Windows with OSMesa
+        if os.name == "nt" and self.windows_skip_image_cache:
+            return
+        # high variation for MacOS
+        if platform.system() == "Darwin" and self.macos_skip_image_cache:
+            return
+
+        # cached image name
+        image_filename = os.path.join(self.cache_dir, test_name[5:] + ".png")
```
MaxJPRey, line 130: @AlejandroFernandezLuces Here, you can add a comment to explain why "[5:]". you could also use

pytest_pyvista.py
```diff
@@ -1,21 +1,168 @@
...
+        error_value=500,
+        warning_value=200,
+        var_error_value=1000,
+        var_warning_value=1000,
+    ):
+        self.test_name = test_name
+
+        if cache_dir is None:
+            # Reset image cache with new images
+            this_path = pathlib.Path(__file__).parent.absolute()
```
MatthewFlamm, line 71: Will this path end up in a site-packages? If so, maybe it should not have a default.

pytest_pyvista.py
```diff
@@ -1,21 +1,168 @@
...
+        if self.reset_image_cache or not os.path.isfile(image_filename):
+            return plotter.screenshot(image_filename)
+
+        # otherwise, compare with the existing cached image
+        error = pyvista.compare_images(image_filename, plotter)
+        if error > allowed_error:
+            raise RuntimeError(
+                f"{test_name} Exceeded image regression error of "
+                f"{allowed_error} with an image error of "
+                f"{error}"
```
MaxJPRey, line 145:
```suggestion
                f"{allowed_error} with an image error equal to: {error}"
```

pytest_pyvista.py
```diff
@@ -1,21 +1,168 @@
...
-    parser.addini('HELLO', 'Dummy pytest.ini setting')
+        self.error_value = error_value
+        self.warning_value = warning_value
+        # self.high_variance_tests = high_variance_tests
+        # self.windows_skip_image_cache = windows_skip_image_cache
+        # self.macos_skip_image_cache = macos_skip_image_cache
+        self.var_error_value = var_error_value
+        self.var_warning_value = var_warning_value
 
+        self.skip = False
```
MaxJPRey, line 87: Do you want to explicitly add the parameter `skip` in the `__init__()` signature?

pytest_pyvista.py
```diff
@@ -1,21 +1,168 @@
...
+    Multiple calls to an instance of this class will append
+    `_X` to the name after the first one.  That is, files
+    ``{test_name}``, ``{test_name}_1``, and ``{test_name}_2``
+    will be saved if called 3 times.
+
+    Parameters
+    ----------
+    test_name : str
+        Name of test to save.  Sets name of image cache file.
+
```
MaxJPRey, line 46: Some parameters such as `cache_dir`, `error_value` are missing in the documentation.

setup.py
```diff
@@ -1,48 +1,48 @@
...
+    author_email="info@pyvista.org",
+    maintainer="The PyVista Developers",
+    maintainer_email="info@pyvista.org",
+    license="MIT",
+    url="https://github.com/pyvista/pytest-pyvista",
+    description="Plugin to test PyVista plot outputs",
+    long_description=read("README.rst"),
+    py_modules=["pytest_pyvista"],
+    python_requires=">=3.7",
+    install_requires=["pytest>=3.5.0"],
```
banesullivan, line 28: Should PyVista be listed as a requirement as well?

pytest_pyvista.py
banesullivan: I'd recommend using `appdirs` for this

MatthewFlamm: This dir is the cached images, so this would be in the repo somewhere likely.  We _could_ allow for the generated images to be stored in a configurable location so that developers could check differences more easily.  I think `appdirs` would be a good default for this second usage.

pytest_pyvista.py
MatthewFlamm: The usage of skip is usually after instantiation since it is instantiated in a fixture in this package.  I assume the user will use it from the fixture typically.  We could add a getter/setter for skip so that it is better documented.

MaxJPRey: @MatthewFlamm Agreed, that would be even better! Yes, this can be done in an other PR as well as the other properties.

AlejandroFernandezLuces: I agree, all attributes of the class should be accessed through getters and setters. I also agree that this should be done in another PR, to keep the size of this one reasonable. I opened an issue regarding this: #4.

setup.py
AlejandroFernandezLuces: Yes, since it is used in the tests. It is added now.

pytest_pyvista.py
AlejandroFernandezLuces: Done. 

`cache_dir` was moved to the pytest command line, which I think will be more useful.

pytest_pyvista.py
AlejandroFernandezLuces: I added the option `--image_cache_dir` to the plugin, so the user can select where the images are saved. 

I tried to also implement `appdirs` as the default. It works fine with `PyVista` tests, but the plugin tests fail with this change because it can't access the `CSIDL_LOCAL_APPDATA` variable from the windows system for some reason.

pytest_pyvista.py
```diff
@@ -1,21 +1,168 @@
...
+            )
+        if error > allowed_warning:
+            warnings.warn(
+                f"{test_name} Exceeded image regression warning of "
+                f"{allowed_warning} with an image error of "
+                f"{error}"
+            )
+
+
+@pytest.fixture(autouse=True)
```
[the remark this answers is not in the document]
AlejandroFernandezLuces: Agree, removed.

pytest_pyvista.py
AlejandroFernandezLuces: Done.

PR #7, opened 2022-11-18
Changes build system to flit

AlejandroFernandezLuces:
### Overview

To standarize the project, I moved the build system to `flit`, which uses the `pyproject.toml` standard. This file allows the project building to be tool independent.


RobPasMue:
``cache-dependency-path: requirements_test.txt`` has to be adapted to use ``pyproject.toml`` ;)

AlejandroFernandezLuces:
I'm not sure how to handle the `cache` option in `actions/setup-python@v4`, since now we are managing all of the project dependencies with `flit`. Currently, I removed it because I don't know if there is any way to keep a cache of the packages with `flit`.

 Is there any option to cache the installed packages using `flit` @RobPasMue ?

.github/workflows/ci_cd.yml
```diff
@@ -41,8 +41,8 @@ jobs:
         uses: actions/setup-python@v4
         with:
           python-version: ${{ matrix.python-version }}
-          cache: 'pip'
-          cache-dependency-path: requirements_test.txt
+          # cache: 'pip'
+          # cache-dependency-path: requirements_test.txt
```
RobPasMue, line 45:
```suggestion
          cache: 'pip'
          cache-dependency-path: pyproject.toml
```

RobPasMue (approved):
I htink that's the only thing you need @AlejandroFernandezLuces :)

.github/workflows/ci_cd.yml
RobPasMue: The way ``cache-dependency-path`` works is that it looks for the hash of a certain file (i.e. it's checksum). If the checksum is modified (i.e. your file has changed) it will update the cache =)

.github/workflows/ci_cd.yml
```diff
@@ -51,11 +51,10 @@ jobs:
...
             Examples-1-
 
-      - name: Build wheel and install pytest-pyvista
+      - name: Build wheel and install pytest-pyvista with test dependencies
         run: |
-          pip install wheel
-          python setup.py bdist_wheel
-          pip install dist/pytest_pyvista*.whl
+          pip install flit
+          flit install --extras tests
```
RobPasMue, line 57:
```suggestion
          pip install .[tests]
```

RobPasMue:
The way ``cache-dependency-path`` works is by checking the checksum of whatever file you provide. If it doesn't change, it will use the stored cache with that hash/checksum. Otherwise, it'll generate a new one. Just keep that in mind.

I also saw that you were installing it using ``flit`` directly. I recommend that you keep using ``pip`` for doing the install process. That way you can make use of the ``cache: 'pip'`` option. I left you another suggestion.

.github/workflows/ci_cd.yml
```diff
@@ -51,11 +51,10 @@ jobs:
           restore-keys: |
             Examples-1-
 
-      - name: Build wheel and install pytest-pyvista
+      - name: Build wheel and install pytest-pyvista with test dependencies
         run: |
-          pip install wheel
-          python setup.py bdist_wheel
-          pip install dist/pytest_pyvista*.whl
+          pip install flit
```
RobPasMue, line 56:
```suggestion
```

MaxJPRey:
Thanks @RobPasMue for looking at the PR. Much appreciated.

.github/workflows/ci_cd.yml
```diff
@@ -51,11 +51,9 @@ jobs:
           restore-keys: |
             Examples-1-
 
-      - name: Build wheel and install pytest-pyvista
+      - name: Build wheel and install pytest-pyvista with test dependencies
```
MaxJPRey, line 54:
```suggestion
      - name: Install pytest-pyvista with test dependencies
```

pyproject.toml
```diff
@@ -0,0 +1,41 @@
...
+tests = [
+    "codecov<2.2.0",
+    "coverage==6.5",
+    "pytest>=3.5.0",
+    "pytest-cov==4.0.0",
+    "pyvista==0.37"
+]
+
+[project.entry-points."pytest11"]
+# Location of the plugin module, in this case ./pytest_pyvista/pytest_pyvista.py
```
MaxJPRey, line 40: This comment might be useful for future maintenance. Thanks for adding it.

MaxJPRey (approved):
LGTM.

Issue #17, opened 2022-12-23
Not creating images for new tests


banesullivan:
The old behavior was that if the image was missing, it would be created. Now you have to use `--reset_image_cache` to generate figures for new tests.

Can we make it so that if the cached image is missing, it gets generated?

MatthewFlamm:
I think that we should store _all_ generated images in a separate dir, defaulted to a temporary dir.  Then we can also compare subtle differences too.

I need to get out of the bad habit of suggesting ideas on other issues that are linked.

PR #19, opened 2022-12-24
Better default cache dir and config options

### LINKED ISSUE #16: Default cache_dir value could be better?
IMO, this default value is not the best choice:

https://github.com/pyvista/pytest-pyvista/blob/374c284d67c962bdb516a77e661f415c30011b47/pytest_pyvista/pytest_pyvista.py#L88-L89

I'd like to request some clarification on this choice and propose altering it to use the current working directory.

I would expect `pytest .`  to properly handle the image cache or default to save the cache in a location that is visible/relative to the active project. This current default dumps the images in a non-visible location.

An alternative option is to have `cache_dir` be easily stable in the `conftest.py` for a project -- is this currently possible?

MatthewFlamm:
Closes #16 

This PR now requires a cache_dir be set within `VeriryImageClass`.  The default is set according to the configuration (`setup.cfg` or `pyproject.toml`) or command line in the fixture.

The existing tests were flawed in that they never tested against an image cache and would always pass.  An image is now dynamically generated for the temporary cache.

Feedback on naming of the public options is specifically requested.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -27,7 +27,7 @@ jobs:
       matrix:
         include:
           - python-version: '3.7'
-            vtk-version: '8.1.2'
+            vtk-version: '9.0.3' # v8.x not supported
           - python-version: '3.8'
             vtk-version: '9.0.3'
           - python-version: '3.9'
--- a/.gitignore
+++ b/.gitignore
@@ -77,3 +77,7 @@ target/
 
 # vscode
 .vscode/
+
+# virtual environment
+venv/
+.venv/
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,7 +33,8 @@ tests = [
     "coverage==6.5",
     "pytest>=3.5.0",
     "pytest-cov==4.0.0",
-    "pyvista>=0.37"
+    "pyvista>=0.37",
+    "numpy<1.24",
 ]
 
 [project.entry-points."pytest11"]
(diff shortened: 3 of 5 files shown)


MatthewFlamm:
We will need to wait for a new release of pyvista to get CI green.

MatthewFlamm:
One thought that I had is that an advanced use case could be separate caches for different use cases, e.g. windows vs Linux.  This PR would in enable this, but the user would have to create their own fixture.

This was my thinking around the dichotomy of requiring a `cache_dir` in the `VerifyImageClass` but then providing a default in the fixture.

banesullivan:
> We will need to wait for a new release of pyvista to get CI green.

Can't we just limit `numpy<1.24` for now?

banesullivan (approved):
Tested locally! Thanks, @MatthewFlamm!!

banesullivan:
This is also solving #17 for me I think - but let's leave that issue open for the other idea you had

MatthewFlamm:
We need this PR to fix testing, but I can't figure out how to easily debug the failing test here on vtk==8.1.2.  To make it easier to debug, it would be very useful to save all generated files into a folder.  But that PR would need these tests as well.  Can we temporarily turn off 8.1.2 check the merge this?  Or I could forge ahead and add this other functionality in this PR?

MatthewFlamm:
As I wrote in #21, we don't even support image regression testing for 8.x.  Or we didn't in pyvista.  For now, let's raise and if there is need to support it, someone can test it and add back in.  I made those changes here since this PR has valid tests and fixed numpy pinning and needs to be merged before any others.

banesullivan:
If I could wave a magic wand, we wouldn't support 8.x at all anywhere so this sounds good to me, thanks for handling all of this!!

PR #21, opened 2022-12-26
Raise for vtk <9

MatthewFlamm:
vtk v8.x is not supported for image regressions.  At least this is how the code works now.  Returning `None` was a workaround in the `pyvista` package instead of skipping the tests.  Now that this package is decoupled, we should raise instead so that users know it isn't supported, otherwise confusion like https://github.com/pyvista/pytest-pyvista/pull/19#issuecomment-1365009955 occurs.


MatthewFlamm:
Closing in favor of #19

PR #22, opened 2022-12-26
Allow saving generated images to dir

### LINKED ISSUE #6: Dump failing images to a directory
I'm dealing with a scenario right now where a lot of tests are failing the image regression test on Windows but not other platforms on CI. It would be nice if there were an option to dump the failing images to a specific path so that I can upload those as a CI artifact to inspect. I'm currently manually doing this

### LINKED ISSUE #17: Not creating images for new tests

MatthewFlamm:
Closes #17 and closes #6. Allow saving generated images to dir.

TODO:
- [x] Add option to append new images to image cache (needed to avoid users moving all new files to image cache)
- [x] ~Reverse option for failing with extra image? IMO, it should be defaulted to fail here and an option to not fail.~ 
 Will leave for a different PR


MatthewFlamm:
This PR is not really a breaking change for adding missing images since this functionality didn't work on the previous releases.  See #17

banesullivan (approved):

pytest_pyvista/pytest_pyvista.py
```diff
@@ -28,6 +27,16 @@ def pytest_addoption(parser):
         action="store_true",
         help="Enables failure if image cache does not exist.",
     )
+    group.addoption(
+        "--generated_image_dir",
+        action="store",
+        help="Path to generated images folder.",
```
banesullivan, line 33:
```suggestion
        help="Path to dump test images from the current run.",
```

tests/test_pyvista.py
```diff
@@ -189,6 +190,49 @@ def test_imcache_var(verify_image_cache):
...
+    )
+
+    result = testdir.runpytest("--fail_extra_image_cache", "--generated_image_dir", "gen_dir")
+    assert os.path.isdir(os.path.join(testdir.tmpdir, "gen_dir"))
+    assert os.path.isfile(os.path.join(testdir.tmpdir, "gen_dir", "imcache.png"))
+    result.stdout.fnmatch_lines("*[Pp]assed*")
+
+
+@skip_vtk8
+def test_gladd_missing_images_commandline(testdir):
```
MatthewFlamm, line 217:
```suggestion
def test_add_missing_images_commandline(testdir):
```

MatthewFlamm:
Merging this after prior approval and merging with upstream with only minor changes afterwards.

PR #23, opened 2022-12-26
Use instance attributes for high variance and skipping based on os

MatthewFlamm:
The current class attributes mean that the values leak out of tests into other tests.  Using instance attributes fixes this problem as the `VerifyImageCache` object is intended to be initialized in a fixture for each test.

Additional tests were added for high variance tests, but not the OS specific skips.


banesullivan (approved):
Thank you for fixing this! This is a significant issue that needs to be merged and released ASAP. I bet you this has been affecting PyVista's tests and as you pointed out elsewhere, might cause testing failures for PyVista

PR #41, opened 2023-02-02
Remove PyVista from dependencies

banesullivan:
Having PyVista as a hard dependency provides a headache for installing targeted versions of both PyVista and VTK.

PyVista has `vtk` listed as a requirement, and I don't think we want to remove that. We are beginning to use new `vtk_osmesa` wheels for CI and we need to be able to have `pytest-pyvista` in `requirements.txt` without a `--no-deps` option. Problem is, this pulls/checks pyvista and thus pulls `vtk`. This provides an issue if using the `vtk_osmesa` (or other build variant) wheel and will also mess with the pyvista install if the version isn't right.

IMO, if you're using `pytest-pyvista`, it's obvious you'll need `pyvista` installed and thus removing it as a dependency here shouldn't be too confusing.

After all, this module is not meant to be directly interfaced with by users but only integrated into existing testing workflows with PyVista.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -32,7 +32,7 @@ jobs:
           cache-dependency-path: pyproject.toml
 
       - name: Install pytest-pyvista with test dependencies
-        run: pip install .[docs]
+        run: pip install .[docs] 'pyvista>=0.37'
 
       - uses: awalsh128/cache-apt-pkgs-action@v1.1.3
         with:
@@ -79,7 +79,7 @@ jobs:
 
       - name: Install pytest-pyvista with test dependencies
         run: |
-          pip install .[tests]
+          pip install .[tests] 'pyvista>=0.37'
 
       - name: Set up vtk
         if: ${{ matrix.vtk-version != 'latest' }}
@@ -108,7 +108,7 @@ jobs:
       github.event_name == 'push' &&
       contains(github.ref, 'refs/tags')
     runs-on: ubuntu-latest
-  
+
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ classifiers=[
     "Operating System :: OS Independent",
     "License :: OSI Approved :: MIT License",
 ]
-dependencies=["pytest>=3.5.0", "pyvista>=0.37"]
+dependencies=["pytest>=3.5.0"]
 python_requires=">=3.7"
 
 [project.urls]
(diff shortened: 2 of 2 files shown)


MatthewFlamm:
I think there is no real harm in removing this as a dependency directly.  As you point out every usage of this tool by definition already requires pyvista to be installed.  But looking at other `pytest-*` packages to gauge user expectations: `pytest-mpl` includes `matplotlib` as a dependency.  `pytest-aiohttp` includes `aiohttp` (by the way this one also has a minimum version). `pytest-freezegun` requires `freezegun` (also with minimum version).  `pytest_httpx` requires `httpx` matching minor version.  If we ever need to pin minimum version for pyvista here (it is probably inevitable), it would be better to start from a state of having a dependency with an unspecified version.

I'd like to see if https://github.com/pyvista/pyvista/pull/3944 could be resolved in a better way before merging this.

banesullivan:
> But looking at other pytest-* packages to gauge user expectations: 

I thought about this as well... and indeed I'm not thrilled to be breaking a pretty clear ecosystem convention.

> I'd like to see if https://github.com/pyvista/pyvista/pull/3944 could be resolved in a better way before merging this.

For posterity, the real issue is: https://github.com/pyvista/pyvista/pull/3704#discussion_r1094927082

AFAIK, there really isn't a good way to handle this since we want to be able to use and `vtk-*` build variant (e.g., `vtk-osmesa`, `vtk-egl`, `vtk-openvr`, etc.) without removing `vtk` from PyVista's core dependencies... which I think we should avoid.

AlejandroFernandezLuces:
I've already had some issues regarding the `pyvista` dependency in this plugin as well. The main issue I see is that previous versions of `pyvista` won't work with this plugin, so maybe some people that need to use earlier versions of `pyvista` won't know which version is compatible with the plugin. It might even lead to people thinking they are using the plugin while they are actually using the test utilities that were used previously in `pyvista`. 

Anyway, if we have to make a tradeoff, I personally prefer to comply with the case you mention, so the plugin can be used in most cases.

AlejandroFernandezLuces (approved):

banesullivan:
@pyvista/developers, any further thoughts?

MatthewFlamm:
Given this:

> AFAIK, there really isn't a good way to handle this since we want to be able to use any vtk-* build variant (e.g., vtk-osmesa, vtk-egl, vtk-openvr, etc.) without removing vtk from PyVista's core dependencies... which I think we should avoid.

If we could have a default for `extras_require` that could be turned off, we could put `vtk` into that and then could be turned off for niche CI usage (and/or the other vtk flavors could be added as non-default options).  But, this didn't exist.

So I'm +1 then.

MatthewFlamm (commented):
LGTM

PR #44, opened 2023-02-14
Remove vtk8 handling

MatthewFlamm:
This PR removes vtk8 handling.

This is related to https://github.com/pyvista/pyvista/pull/4018.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -60,7 +60,7 @@ jobs:
       matrix:
         include:
           - python-version: '3.7'
-            vtk-version: '8.1.2'
+            vtk-version: '9.0.3'
           - python-version: '3.8'
             vtk-version: '9.0.3'
           - python-version: '3.9'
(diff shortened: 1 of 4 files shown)


MatthewFlamm:
I recommend waiting until https://github.com/pyvista/pyvista/pull/3938 is merged before also dropping python 3.7 support here.  It will be easier overall.

.github/workflows/ci_cd.yml
```diff
@@ -60,7 +60,7 @@ jobs:
       matrix:
         include:
           - python-version: '3.7'
-            vtk-version: '8.1.2'
+            vtk-version: '9.0.3'
           - python-version: '3.8'
             vtk-version: '9.0.3'
```
banesullivan, line 65: Can we bump the VTK version on each Python version so as to test against a different VTK version in each job

MatthewFlamm: I have decided to instead align the versions with https://github.com/pyvista/pyvista/pull/4018 by adding in python 3.11 support.  While this doubles up 9.0.3 on 3.7 and 3.8, this is only temporary until 3.7 support is dropped in both packages.

banesullivan: Ah, good call, thanks

banesullivan (approved):

MatthewFlamm:
Shall we have a new patch release for this?  At some point we will have to start doing minor releases, but this package still feels in the alpha/early-beta stage.

PR #49, opened 2023-03-08
Add downstream tests and guard against leakages across tests

### LINKED ISSUE #43: Add pyvista tests into CI
It would be useful to run the full pyvista tests on one python/vtk version in this package to avoid unintended breakages downstream.

### LINKED ISSUE #50: Teardown verify_image_cache after each test
See #49 for some effects where it seems that `verify_image_cache` is leaking outside of test scope.

We should teardown, e.g. removing the object from the theme, to reduce cross talk, after each test.

MatthewFlamm:
closes #43 and closes #50

### DIFF
--- a/tests/test_pyvista.py
+++ b/tests/test_pyvista.py
@@ -251,3 +251,33 @@ def test_reset_image_cache(testdir):
     assert not filecmp.cmp(filename, filename_original, shallow=False)
     # should pass even if image doesn't match
     result.stdout.fnmatch_lines("*[Pp]assed*")
+
+
+def test_cleanup(testdir):
+    """Test cleanup of the `verify_image_cache` fixture."""
+    make_cached_images(testdir.tmpdir)
+    testdir.makepyfile(
+        """
+       import pytest
+       import pyvista as pv
+       pv.OFF_SCREEN = True
+
+       @pytest.fixture()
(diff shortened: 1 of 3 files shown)


tests/test_pyvista.py
```diff
@@ -251,3 +251,33 @@ def test_imcache(verify_image_cache):
...
+       import pytest
+       import pyvista as pv
+       pv.OFF_SCREEN = True
+
+       @pytest.fixture()
+       def cleanup_tester():
+           yield
+           assert pv.global_theme.before_close_callback is None
+
+       def test_imcache(cleanup_tester, verify_image_cache):
```
MatthewFlamm, line 270: The order of fixtures, `a` and `b`, in ```def test_func(a, b):``` is left to right, then after `test_imcache` it runs right to left as expected from multiple yielding.  So `a` fixture is called first, which yields, then `b` fixture is called next, which also yields.  After `test_func` is run, `b` is finished, then `a` is finished.  By putting `cleanup_tester` first in the order, the `assert` is line is executed last.

This also works the same with `request.addfinalizer` in my testing.

akaszynski:
@MatthewFlamm, we ready for another patch release?

MatthewFlamm:
See #53

PR #51, opened 2023-03-09
Bug: guard against leaking verify_image_cache

MatthewFlamm:
Note the base branch is from #49, which has the failing tests.

Will address #50, but won't close it until it is merged into main.

### DIFF
--- a/pytest_pyvista/pytest_pyvista.py
+++ b/pytest_pyvista/pytest_pyvista.py
@@ -241,4 +241,6 @@ def verify_image_cache(request, pytestconfig):
     )
     pyvista.global_theme.before_close_callback = verify_image_cache
 
-    return verify_image_cache
+    yield verify_image_cache
+    
+    pyvista.global_theme.before_close_callback = veryify_image_cache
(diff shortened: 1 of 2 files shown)


pytest_pyvista/pytest_pyvista.py
```diff
@@ -241,4 +245,9 @@ def verify_image_cache(request, pytestconfig):
     )
     pyvista.global_theme.before_close_callback = verify_image_cache
 
+    def reset():
+
```
MatthewFlamm, line 249:
```suggestion
```

pytest_pyvista/pytest_pyvista.py
```diff
@@ -241,4 +245,8 @@ def verify_image_cache(request, pytestconfig):
     )
     pyvista.global_theme.before_close_callback = verify_image_cache
 
+    def reset():
+        pyvista.global_theme.before_close_callback = None
```
MatthewFlamm, line 249: I'm wondering if we should additionally do `pyvista.close_all` in `reset` to help downstream libs handle cleanup.  However, it might give confusing results if they use this fixture in some tests but not others.  Maybe this warrants a separate fixture here like `auto_close`.  We could use a flag to turn it on and off.  This should be a separate PR as it adds new features if implemented this way.

PR #66, opened 2023-08-15
Fix CI; Remove python 3.7; remove flake8-isort

MatthewFlamm:
CI is starting to fail due to conflicts with python 3.7. PyVista no longer supports this version, so we should drop here too.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -59,8 +59,6 @@ jobs:
 
       matrix:
         include:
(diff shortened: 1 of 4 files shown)


MatthewFlamm:
flake8-isort is broken due to incompatibility with the newest version of flake8.  We already have isort so it is unneeded and flake8-isort has an unfavorable license.  Finally isort needed to be updated to fix a poetry installation issue.

akaszynski (approved):

PR #67, opened 2023-08-23
pin sphinx

MatthewFlamm:


MatthewFlamm:
All other PRs are now failing due to sphinx incompatibility.  We should pin so we can control this.

akaszynski (approved):

PR #77, opened 2023-09-14
Change dependency/CI version updates to monthly

MatthewFlamm:
Daily or whatever the frequency of updates is too frequent for this package.


MatthewFlamm:
Also uses the same settings for pre-commit configuration as pyvista/pyvista.

akaszynski (approved):
Agreed. Pre-commit can be quarterly as well to reduce the spam.

PR #80, opened 2023-09-20
Adding custom error class

### LINKED ISSUE #79: Using a custom error class
This is useful when used together with [pytest-rerunfailures](https://github.com/pytest-dev/pytest-rerunfailures), and it can help to also with `try` context managers.

germa89:
As the title.

Close #79

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -102,7 +102,7 @@ jobs:
 
       - name: Unit Testing
         run: xvfb-run coverage run -m --source=pytest_pyvista --module pytest --verbose .
-      - uses: codecov/codecov-action@v3
+      - uses: codecov/codecov-action@v4
         if: matrix.python-version == '3.9'
         name: 'Upload coverage to CodeCov'
 
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -102,7 +102,7 @@ jobs:
 
       - name: Unit Testing
         run: xvfb-run coverage run -m --source=pytest_pyvista --module pytest --verbose .
-      - uses: codecov/codecov-action@v4
+      - uses: codecov/codecov-action@v3
         if: matrix.python-version == '3.9'
         name: 'Upload coverage to CodeCov'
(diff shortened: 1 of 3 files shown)


germa89:
pre-commit.ci autofix

.github/workflows/ci_cd.yml
```diff
@@ -102,7 +102,7 @@ jobs:
 
       - name: Unit Testing
         run: xvfb-run coverage run -m --source=pytest_pyvista --module pytest --verbose .
-      - uses: codecov/codecov-action@v4
+      - uses: codecov/codecov-action@v3
```
MatthewFlamm, line 105:
```suggestion
      - uses: codecov/codecov-action@v4
```

pytest_pyvista/pytest_pyvista.py
```diff
@@ -8,6 +8,20 @@
 import pyvista
 
 
+class RegressionError(RuntimeError):
+    """Error when regression does not meet the criteria"""
+
+    def __init__(self, msg):
```
MatthewFlamm, line 14: Do we need these __init__ methods?  Can't we instead inherit the parent class method?

MatthewFlamm (commented):
Looks good to me. Please use in the tests here too, I think there is only one spot currently.  Also, can we add tests that these get raised?

.github/workflows/ci_cd.yml
germa89: I think it was deprecated.

pytest_pyvista/pytest_pyvista.py
germa89: Maybe you don't needed. I guess it is just an habit.

.github/workflows/ci_cd.yml
germa89: See https://github.com/pyvista/pytest-pyvista/actions/runs/6254763049/job/16982901626

germa89:
I think I'm done with this!

MatthewFlamm (approved):
LGTM

PR #82, opened 2023-09-25
Reset cache only if error

### LINKED ISSUE #83: Reset cache only if error
I think this is a very specific need, but it might be useful to somebody else.

I would like to generate the cache only for the images that fails.

germa89:
As the title.

It resets the cache if the image comparison fails.
It does not raise an error.

Working on the tests...

Close #83

### DIFF
--- a/pytest_pyvista/pytest_pyvista.py
+++ b/pytest_pyvista/pytest_pyvista.py
@@ -108,6 +108,7 @@ class VerifyImageCache:
     ignore_image_cache = False
     fail_extra_image_cache = False
     add_missing_images = False
+    reset_only_failed = False
 
     def __init__(
         self,
@@ -118,8 +119,7 @@ class VerifyImageCache:
         warning_value=200.0,
         var_error_value=1000.0,
         var_warning_value=1000.0,
-        generated_image_dir=None,
-        reset_only_failed=False,
+        generated_image_dir=None
     ):
         self.test_name = test_name
 
@@ -135,7 +135,6 @@ class VerifyImageCache:
         self.warning_value = warning_value
         self.var_error_value = var_error_value
         self.var_warning_value = var_warning_value
-        self.reset_only_failed = reset_only_failed
 
         self.generated_image_dir = generated_image_dir
         if self.generated_image_dir is not None and not os.path.isdir(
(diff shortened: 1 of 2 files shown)


pytest_pyvista/pytest_pyvista.py
```diff
@@ -217,7 +217,12 @@ def __call__(self, plotter):
         error = pyvista.compare_images(image_filename, plotter)
 
         if error > allowed_error:
+            print("self.reset_only_failed", self.reset_only_failed)
```
germa89, line 220: Shouldn't this be deleted?

pytest_pyvista/pytest_pyvista.py
```diff
@@ -245,6 +250,7 @@ def verify_image_cache(request, pytestconfig):
         "fail_extra_image_cache"
     )
     VerifyImageCache.add_missing_images = pytestconfig.getoption("add_missing_images")
+    VerifyImageCache.reset_only_failed = pytestconfig.getoption("reset_only_failed")
```
germa89, line 254: Good catch!

pytest_pyvista/pytest_pyvista.py
AlejandroFernandezLuces: Yes, missed it

tests/test_pyvista.py
```diff
@@ -281,3 +281,23 @@ def test_imcache(cleanup_tester, verify_image_cache):
...
+        def test_imcache(verify_image_cache):
+            sphere = pv.Box()
+            plotter = pv.Plotter()
+            plotter.add_mesh(sphere, color="blue")
+            plotter.show()
+        """
+    )
+
+    result = testdir.runpytest("--reset_only_failed")
+    result.stdout.fnmatch_lines("*[Pp]assed*")
```
MatthewFlamm, line 302: We should also test that the file was indeed overwritten, see https://github.com/pyvista/pytest-pyvista/blob/42f771b84fd99d98a1e202f432ee56a9773e8f05/tests/test_pyvista.py#L251

pytest_pyvista/pytest_pyvista.py
```diff
@@ -197,23 +205,32 @@ def __call__(self, plotter):
             self.add_missing_images
             and not os.path.isfile(image_filename)
             or self.reset_image_cache
-        ):
+        ) and not self.reset_only_failed:
```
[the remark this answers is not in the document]
MatthewFlamm: Oh I see `--reset_ONLY_failed`.  This is fine with me then.  Sorry for the noise.

tests/test_pyvista.py
AlejandroFernandezLuces: Agree, resolved.

MatthewFlamm (approved):
LGTM

PR #84, opened 2023-09-26
Update README with conda/mamba install

bjlittle:
This pull-request updates the `README.rst` with simple instructions to the user community on how to install `pytest-pyvista` using `conda` or `mamba` from the newly hatched package on [conda-forge](https://anaconda.org/conda-forge/pytest-pyvista).

PR #85, opened 2023-09-26
NEP29 Compliance

bjlittle:
This PR adopts the [NEP29 Drop Schedule](https://numpy.org/neps/nep-0029-deprecation_policy.html#drop-schedule) for minimum supported versions of `python` and `numpy`.

When working on #84 I noticed that the PyPI versions were wrong for `pytest-pyvista`. Turns out, incorrect syntax (probably migrated from `setup.py` or `setup.cfg`) was used to specify the required `python` version for the package; minor fix sorted this.

Went the extra mile and also configured the  [validate-pyproject](https://github.com/abravalheri/validate-pyproject) pre-commit hook to automate continued confidence in the `pyproject.toml`.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -59,8 +59,6 @@ jobs:
 
       matrix:
         include:
-          - python-version: '3.8'
-            vtk-version: '9.0.3'
           - python-version: '3.9'
             vtk-version: '9.1'
           - python-version: '3.10'
@@ -84,10 +82,6 @@ jobs:
       - name: Set up vtk
         if: ${{ matrix.vtk-version != 'latest' }}
         run: pip install vtk==${{ matrix.vtk-version }}
-
-      - name: Limit NumPy for VTK 9.0.3
-        if: ${{ matrix.vtk-version == '9.0.3' }}
-        run: pip install 'numpy<1.24'
 
       - uses: awalsh128/cache-apt-pkgs-action@v1.1.3
         with:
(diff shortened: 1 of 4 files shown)


tkoyama010 (approved):
Thanks! I wondered the same thing. About adding `validate-pyproject`, also ask the @MatthewFlamm if it would be useful as it would increase run time. If it is adopted here, please consider sending an additional PR to `pyvista/pyvista`.

bjlittle:
> Thanks! I wondered the same thing. About adding `validate-pyproject`, also need to ask the @MatthewFlamm if it would be useful to add as it would increase run time. If it is adopted here, please consider sending an additional PR to [pyvista/pyvista](https://github.com/pyvista/pyvista) too.

@tkoyama010 Cool, thanks.

Although, it seems to be a relatively snappy pre-commit hook. I'm not seeing much difference in runtime or resources being used.

MatthewFlamm:
Since this package is required for testing pyvista, we can't drop python 3.8 support without dropping it in PyVista.  We should probably move the discussion there.

tkoyama010:
Oh, I see. I totally forgot this project was used in PyVista😅.

tkoyama010:
I started https://github.com/pyvista/pyvista/discussions/4960 . Let's discuss it there!

tkoyama010 (changes requested):
Keep this while we discuss it.

bjlittle:
@MatthewFlamm and @tkoyama010 

Whilst we await a decision on this one, I'm going to put the pull-request into `draft` to ensure it's not merged 👍

akaszynski:
Out of date, dropping. Thanks for pointing this out @bjlittle.

PR #92, opened 2023-10-02
Renaming CLI argument

germa89:
Renaming `fail_extra_image_cache` argument to `fail_if_missing_image_cache`, since this new name is more self-explanatory. It might be a bit too long. I'm happy to delete the `_if` part if needed.


akaszynski:
@MatthewFlamm, it's your call.

MatthewFlamm:
I'm +1.  I assume there are only a few downstream users of this package.  Should we deprecate the old one and make it a temporary alias?  I'm fine just hard removing it too given how young this package is.

germa89:
I would just remove it...

MaxJPRey:
I would just remove it too. I will notify people making usage of it in PyAnsys packages.

MatthewFlamm:
This was again identified as a misunderstanding in https://github.com/pyvista/pyvista/issues/7582.  Should this still be merged and a new release issued?

MatthewFlamm:
I also can't think of a good reason that this should be off by default.  If changed behavior, the flag could be ``--pass-if-missing-image``.

MatthewFlamm:
We have three thumbs up (counting the one to https://github.com/pyvista/pytest-pyvista/pull/92#issuecomment-2935835665) then to change the default behavior.  I like your suggestion about

> Or even:
> --allow_unused_generated
> --allow_unused_cache

MatthewFlamm:
Superceded by #175

PR #143, opened 2024-11-01
Drop Python 3.8 support which is EOL in 07 Oct 2024

tkoyama010:
ref https://github.com/pyvista/pyvista/pull/6701


akaszynski (approved):
LGTM

PR #169, opened 2025-06-14
Add `disallow_unused_cache` option

### LINKED ISSUE #28: Add option to check if a cached image isn't compared to an image during all tests
It would be very useful to have an option to check that all cached images were compared during a test.  This is a complement to checking that all comparisons have an image in the cache.

This is tricky if tests are skipped for any reason.  So, it would be only useful if there is a canonical version for which no test is skipped.  I'm not sure if this is the case today for `pyvista`.  It is very possible that at least one plotting test is skipped on every vtk version.  But, this could be useful for downstream libs that have less wide-open vtk version support.

user27182:
New CLI option to fix #28 and https://github.com/pyvista/pyvista/issues/7582

This works by storing test results info into ~a global dict~ global sets. The ~dict~ set is populated each time `verify_image_cache` is called, OR whenever a test is skipped during test setup or test execution. This should handle cases where tests are skipped based on OS version or VTK version and not inadvertently fail.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -174,7 +174,7 @@ jobs:
         working-directory: pyvista
 
       - name: Unit Testing
-        run: xvfb-run python -m pytest -v --disallow_unused_cache --generated_image_dir gen_dir tests/plotting/test_plotting.py
+        run: xvfb-run python -m pytest -v --generated_image_dir gen_dir tests/plotting/test_plotting.py
         working-directory: pyvista
 
       - name: Upload generated image artifact
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -174,7 +174,7 @@ jobs:
         working-directory: pyvista
 
       - name: Unit Testing
-        run: xvfb-run python -m pytest --fail_extra_image_cache -v --generated_image_dir gen_dir tests/plotting/test_plotting.py
+        run: xvfb-run python -m pytest --fail_extra_image_cache -v --generated_image_dir gen_dir tests/plotting/test_plotting.py --allow_unused_cache
         working-directory: pyvista
 
       - name: Upload generated image artifact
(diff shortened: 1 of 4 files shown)


tkoyama010 (approved):
LGTM

MatthewFlamm:
Before we merge some of these, I wonder whether we should make the default behavior better and make this one and #92 turned on by default. We'd have to rename the options.

user27182:
> Before we merge some of these, I wonder whether we should make the default behavior better and make this one and #92 turned on by default. We'd have to rename the options.

Agreed, let's standardize the default values and names as of part of the discussion for #92
We should also consider enabling #170 by default
And, this PR has a global `RESULTS` dict, which may not play nicely with parallel testing https://github.com/pyvista/pyvista/pull/7624, so we should hold off for that too

I think #171 can probably be merged as-is though? It has no parallel issues and the option is already enabled by default

user27182:
> Before we merge some of these, I wonder whether we should make the default behavior better and make this one and #92 turned on by default. We'd have to rename the options.

Turning on the option from this PR by default may not be a good idea after all. If we always fail tests when a cached image is not used, then running pytest for anything other than the full test suite will always result in failure. E.g. debugging with `pytest -k legend` will fail since only the legend images are used, and hence all other cached images are unused and so the tests fail. Similarly, simply doing `pytest --doctest-modules` will also fail since all cached images will be unused.

So while it makes sense to raise an error by default if a _generated_ image is not used, it might not make sense to do the same (by default) if a _cached_ image is not used

user27182:
Downstream tests with https://github.com/pyvista/pyvista/pull/7616 look good

MatthewFlamm:
I won't have time to look at this PR closely in the next week or so, but the naming is good to me and the default being off makes sense to me too.

akaszynski:
@user27182, looks like I screwed up the merge. Would you mind taking a quick look at this so we can get a release out this weekend?

user27182:
I thinks there's a strange interaction happening with `pytest_runtest_makereport` added from #170 and this PR

Might be a pytest thing... will have to look into it

akaszynski:
Seems to be working. Ping me when it's out of draft and ready for review.

user27182:
@akaszynski this is good to go

PR #170, opened 2025-06-14
Add `allow_useless_fixture` CLI option

user27182:
The `verify_image_cache` fixture is sometimes used unnecessarily when no calls to `plot` or show` are made and hence there is nothing to verify. This PR adds an option to check for this and will fail the test's teardown.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -174,7 +174,7 @@ jobs:
         working-directory: pyvista
 
       - name: Unit Testing
-        run: xvfb-run python -m pytest --fail_extra_image_cache -v --generated_image_dir gen_dir tests/plotting/test_plotting.py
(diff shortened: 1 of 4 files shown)


user27182:
Downstream tests with https://github.com/pyvista/pyvista/pull/7617 look good

user27182:
@akaszynski are you interested in reviewing this? If we get this PR, #169 , and maybe #176 merged, then we can consider making a new release so that we can use these new options downstream (we have three new CLI options for use by pyvista with https://github.com/pyvista/pyvista/pull/7616, https://github.com/pyvista/pyvista/pull/7617, https://github.com/pyvista/pyvista/pull/7620)

akaszynski (approved):
Thanks for adding this. LGTM, I see no issues. Agreed regarding a new release, we can get that out this weekend.

PR #171, opened 2025-06-15
Add `failed_image_dir` option

user27182:
New option to save any tests where a warning or error was generated. 

Currently, if a test fails and we want to review the results, we have to manually search for the generated image in the `generated_image_dir` as well as manually search for the cached image in the `cache_image_dir`. This can be challenging with 500+ images in the pyvista repo. With this PR, these results are captured and copied into a separate easy-to-review location in `failed_image_dir`. This is similar to how the documentation image tests currently work for pyvista.

~This PR builds upon code from #169~

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -108,9 +108,8 @@ These are the flags you can use when calling ``pytest`` in the command line:
   directory.  This will override any configuration, see below.
 
 * ``--failed_image_dir <DIR>`` dumps copies of cached and generated test images when
-  there is a warning or error raised. Regression error values are also summarized in
-  a JSON file. This directory is useful for reviewing test failures.
-  This will override any configuration, see below.
+  there is a warning or error raised. This directory is useful for reviewing test
+  failures. This will override any configuration, see below.
 
 * ``--add_missing_images`` adds any missing images from the test run to the cache.
 
--- a/README.rst
+++ b/README.rst
@@ -157,6 +157,12 @@ Additionally, to configure the directory that will contain the generated test im
    [tool.pytest.ini_options]
    generated_image_dir = "generated_images"
 
+Similarly, configure the directory that will contain any failed test images:
+
+.. code::
+
+   [tool.pytest.ini_options]
+   failed_image_dir = "failed_images"
 
 Contributing
 ------------
--- a/README.rst
+++ b/README.rst
@@ -108,8 +108,9 @@ These are the flags you can use when calling ``pytest`` in the command line:
   directory.  This will override any configuration, see below.
 
 * ``--failed_image_dir <DIR>`` dumps copies of cached and generated test images when
-  there is a warning or error raised. This directory is useful for reviewing test
-  failures. This will override any configuration, see below.
+  there is a warning or error raised. Regression error values are also summarized in
+  a JSON file. This directory is useful for reviewing test failures.
+  This will override any configuration, see below.
 
 * ``--add_missing_images`` adds any missing images from the test run to the cache.
 
--- a/README.rst
+++ b/README.rst
@@ -114,6 +114,9 @@ These are the flags you can use when calling ``pytest`` in the command line:
 
 * ``--reset_only_failed`` reset the image cache of the failed tests only.
 
+* ``--fail_unused_cache`` report test failure if there are any images in the cache
+  which are not compared to any generated images.
+
 Test specific flags
 -------------------
 These are attributes of `verify_image_cache`. You can set them as ``True`` if needed
--- a/README.rst
+++ b/README.rst
@@ -107,15 +107,16 @@ These are the flags you can use when calling ``pytest`` in the command line:
 * ``--generated_image_dir <DIR>`` dumps all generated test images into the provided
   directory.  This will override any configuration, see below.
 
+* ``--failed_image_dir <DIR>`` dumps copies of cached and generated test images when
+  there is a warning or error raised. This directory is useful for reviewing test
+  failures. This will override any configuration, see below.
+
 * ``--add_missing_images`` adds any missing images from the test run to the cache.
 
 * ``--image_cache_dir <DIR>`` sets the image cache directory.  This will override any
   configuration, see below.
 
 * ``--reset_only_failed`` reset the image cache of the failed tests only.
-
-* ``--fail_unused_cache`` report test failure if there are any images in the cache
-  which are not compared to any generated images.
 
 Test specific flags
 -------------------
--- a/pytest_pyvista/pytest_pyvista.py
+++ b/pytest_pyvista/pytest_pyvista.py
@@ -284,10 +284,10 @@ def _ensure_dir_exists(dirpath: str, msg_name: str) -> None:
         Path(dirpath).mkdir()
 
 
-def _get_option_from_config_or_ini(pytestconfig, dirname: str) -> str:  # noqa: ANN001
-    gen_dir = pytestconfig.getoption(dirname)
+def _get_option_from_config_or_ini(pytestconfig, option: str) -> str:  # noqa: ANN001
+    gen_dir = pytestconfig.getoption(option)
     if gen_dir is None:
-        gen_dir = pytestconfig.getini(dirname)
+        gen_dir = pytestconfig.getini(option)
     return gen_dir
 
 
--- a/pytest_pyvista/pytest_pyvista.py
+++ b/pytest_pyvista/pytest_pyvista.py
@@ -315,7 +315,8 @@ def _combine_temp_jsons(parent_dir: Path) -> None:
                 json.dump(dict(sorted(combined_data.items())), f, indent=2)
 
         # Remove tmp dir
-        shutil.rmtree(tmp_jsons_dir)
+        if tmp_jsons_dir.exists():
+            shutil.rmtree(tmp_jsons_dir)
 
 
 @pytest.fixture
--- a/pytest_pyvista/pytest_pyvista.py
+++ b/pytest_pyvista/pytest_pyvista.py
@@ -2,21 +2,16 @@
 
 from __future__ import annotations
 
-from enum import Enum
 import os
 from pathlib import Path
 import platform
 import shutil
-import sys
-from typing import TYPE_CHECKING
-from typing import NamedTuple
+from typing import Literal
+from typing import cast
 import warnings
 
 import pytest
 import pyvista
-
-if TYPE_CHECKING:  # pragma: no cover
-    from collections.abc import Generator
 
 
 class RegressionError(RuntimeError):
@@ -25,15 +20,6 @@ class RegressionError(RuntimeError):
 
 class RegressionFileNotFound(FileNotFoundError):  # noqa: N818
     """Error when regression file is not found."""
-
-
-class Outcome(Enum):
-    """Outcome of the image verification."""
-
-    SUCCESS = "success"
-    SKIPPED = "skipped"
-    ERROR = "error"
-    WARNING = "warning"
 
 
 def pytest_addoption(parser) -> None:  # noqa: ANN001
@@ -89,11 +75,6 @@ def pytest_addoption(parser) -> None:  # noqa: ANN001
         "--reset_only_failed",
         action="store_true",
         help="Reset only the failed images in the PyVista cache.",
-    )
-    group.addoption(
-        "--fail_unused_cache",
-        action="store_true",
-        help="Report test failure if there are any images in the cache which are not compared to any generated images.",
     )
 
 
@@ -166,6 +147,7 @@ class VerifyImageCache:
         var_error_value=1000.0,  # noqa: ANN001
         var_warning_value=1000.0,  # noqa: ANN001
         generated_image_dir=None,  # noqa: ANN001
+        failed_image_dir=None,  # noqa: ANN001
     ) -> None:
         self.test_name = test_name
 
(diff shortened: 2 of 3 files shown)


user27182:
Ready but the downstream tests are failing because of pyvista's  `--strict` pytest flag, which converts the warning from this PR about `failed_test_dir` not existing into an error. https://github.com/pyvista/pyvista/pull/7619 will need to be merged first.

pytest_pyvista/pytest_pyvista.py
```diff
@@ -239,19 +329,32 @@ def verify_image_cache(request, pytestconfig):  # noqa: ANN001, ANN201
...
     def reset() -> None:
         pyvista.global_theme.before_close_callback = None
 
     request.addfinalizer(reset)  # noqa: PT021
     return verify_image_cache
+
+
+@pytest.hookimpl
+def pytest_sessionfinish(session, exitstatus) -> None:  # noqa: ANN001, ARG001
+    """Merge and clean up temp json files."""
```
user27182, line 351: This is needed because we wrote individual JSON files for each test result earlier. This is done because writing to a single JSON file or updating a single global dict variable would presumably create race conditions and break usage of things like `pytest-xdist`. This adds a bit of complexity, see changes from commit https://github.com/pyvista/pytest-pyvista/pull/171/commits/176279477466b7a9327f9ef33d7439fe085d78ad. We can revert if parallel processing isn't an issue...

user27182: Saving the error to JSON is kind of extra and not really needed since the errors are already summarized by pytest. So this feature is completely removed for simplicity.

pytest_pyvista/pytest_pyvista.py
```diff
@@ -209,10 +224,18 @@ def __call__(self, plotter):  # noqa: ANN001, ANN204, C901, PLR0912
         if self.generated_image_dir is not None:
             gen_image_filename = os.path.join(self.generated_image_dir, test_name[5:] + ".png")  # noqa: PTH118
             plotter.screenshot(gen_image_filename)
+            if not Path(image_filename).is_file():
+                # Image comparison will fail, so save image before error
+                self._save_failed_test_images("error", plotter, image_name)
+
+                # Make sure this doesn't get called again
+                plotter._before_close_callback = None  # noqa: SLF001
```
MatthewFlamm, line 232: I _think_ this line above was needed because there was an error raised in this code which then causes the plotter to not close properly and leading to this code being called again.  Since there is no error raised here, why is this line here?

pytest_pyvista/pytest_pyvista.py
```diff
@@ -226,8 +249,45 @@ def __call__(self, plotter):  # noqa: ANN001, ANN204, C901, PLR0912
...
+            _ensure_dir_exists(self.failed_image_dir, msg_name="failed image dir")
+            Path(self.failed_image_dir, errors_or_warnings).mkdir(exist_ok=True)
+
+            dest_dir = Path(self.failed_image_dir, errors_or_warnings, from_cache_or_test)
+            dest_dir.mkdir(exist_ok=True)
+            return dest_dir
+
+        error_dirname = cast("Literal['errors', 'warnings']", error_or_warning + "s")
+        from_dir = _make_failed_test_image_dir(error_dirname, "from_test")
+        plotter.screenshot(Path(from_dir, image_name))
```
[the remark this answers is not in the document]
user27182: I'm not sure if calling `screenshot` is expensive though, and keeping track of the files and doing `is_file` checks and `shutil.copy` may be a pain since the screenshot is only conditionally generated. But, it would help ensure the screenshot is always the same. Instead of managing the files, we could refactor `plotter.screenshot(image_filename)` as a separate function:

``` python
def save_generated_image():
    plotter.screenshot(image_filename)
```

That way we ensure that the call is always the same at least

pytest_pyvista/pytest_pyvista.py
```diff
@@ -226,8 +249,45 @@ def __call__(self, plotter):  # noqa: ANN001, ANN204, C901, PLR0912
...
+
+
+def _ensure_dir_exists(dirpath: str, msg_name: str) -> None:
+    if not Path(dirpath).is_dir():
+        msg = f"pyvista test {msg_name}: {dirpath} does not yet exist.  Creating dir."
+        warnings.warn(msg, stacklevel=2)
+        Path(dirpath).mkdir()
+
+
+def _get_dir_from_config_or_ini(pytestconfig, dirname: str) -> str:  # noqa: ANN001
```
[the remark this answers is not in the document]
user27182: done in https://github.com/pyvista/pytest-pyvista/pull/171/commits/8ae2a8648b418636cc73ba449131a4938aabe772

README.rst
```diff
@@ -107,6 +107,10 @@ These are the flags you can use when calling ``pytest`` in the command line:
 * ``--generated_image_dir <DIR>`` dumps all generated test images into the provided
   directory.  This will override any configuration, see below.
 
+* ``--failed_image_dir <DIR>`` dumps copies of cached and generated test images when
```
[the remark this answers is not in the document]
user27182: done in https://github.com/pyvista/pytest-pyvista/pull/171/commits/24c55d2dec4630dd319874967674c394b8dff392

pytest_pyvista/pytest_pyvista.py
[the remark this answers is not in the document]
user27182: I removed the use of `plotter.screenshot` from this PR in favor of copying the image here https://github.com/pyvista/pytest-pyvista/pull/171/commits/90f99161c65c3411dd356a377e4e440689f609a9

pytest_pyvista/pytest_pyvista.py
```diff
@@ -221,13 +247,56 @@ def __call__(self, plotter):  # noqa: ANN001, ANN204, C901, PLR0912
...
+
+        error_dirname = cast("Literal['errors', 'warnings']", error_or_warning + "s")
+
+        if self.generated_image_dir is not None:
+            generated_image = Path(self.generated_image_dir, image_name)
+            if generated_image.is_file():
+                from_test_dir = _make_failed_test_image_dir(error_dirname, "from_test")
+                shutil.copy(generated_image, Path(from_test_dir, image_name))
+
+        if self.cache_dir is not None:
```
MatthewFlamm, line 280: Can this actually be `None`?  How would `verify_image_cache` work at all in this case?  Is this another bug or unintended feature?  Again maybe a separate PR.

pytest_pyvista/pytest_pyvista.py
```diff
@@ -221,13 +247,56 @@ def __call__(self, plotter):  # noqa: ANN001, ANN204, C901, PLR0912
...
+            errors_or_warnings_dir = Path(self.failed_image_dir, errors_or_warnings)
+            errors_or_warnings_dir.mkdir(exist_ok=True)
+
+            dest_dir = errors_or_warnings_dir / from_cache_or_test
+            dest_dir.mkdir(exist_ok=True)
+            return dest_dir
+
+        error_dirname = cast("Literal['errors', 'warnings']", error_or_warning + "s")
+
+        if self.generated_image_dir is not None:
```
MatthewFlamm, line 274: It is likely my previous comment spurred this, but now we cannot generate _only_ the failed images.

My original thought is that we were generating the image for the comparison itself, and then we could move it around to the various locations as needed.  If this were possible, it would greatly simplify all the logic (not just for this feature).  However, this is not what happens.  `pyvista.compare_images` works differently than `Plotter.screenshot`.  This raises questions about the `1:1` nature of `generated_image_dir` to the actual image that is compared to the cache.  I wonder if in the future, we should instead create an actual figure using `screenshot` in a tmp filespace and then move it around to ensure 1:1 mapping.

pytest_pyvista/pytest_pyvista.py
user27182: I am not actually sure. If it's possible, it should be an error. I added this because I saw this check is also done for `self.generated_image_dir` elsewhere, so added it here for completeness.

We should enable branch coverage with `--cov-branch` for codecov, this will indicate missing coverage if this branch evaluates to False, which is the case we'd need to test for

pytest_pyvista/pytest_pyvista.py
user27182: > It is likely my previous comment spurred this, but now we cannot generate only the failed images.

Ah yes I see what you're saying. We did `plotter.screenshot` earlier, which we could do regardless if the generated image was saved on not. But now I've made the `failed_image_dir` option conditional on using the `generated_image_dir` option too, we don't want.

> My original thought is that we were generating the image for the comparison itself, and then we could move it around to the various locations as needed. If this were possible, it would greatly simplify all the logic (not just for this feature).

This was also my understanding, which led me to change this function so it copies the generated image instead of saving a new screenshot... but I overlooked the dependency above.

> However, this is not what happens.  pyvista.compare_images works differently than Plotter.screenshot. This raises questions about the 1:1 nature of generated_image_dir to the actual image that is compared to the cache.

I am not sure I follow what you mean, but I thought `screenshot` saves an image as png, and `compare_images` reads and compares the `png` files, so are they not using the same image?

> I wonder if in the future, we should instead create an actual figure using screenshot in a tmp filespace and then move it around to ensure 1:1 mapping.

Yes we could do this, OR, use `filename=None` and store the image in memory directly with `return_img=True`, and only save if `generated_image_dir` is not None and/or if `failed_image_dir` is not None. Then we don't even need to move it around, but only save it as needed. TBH though I think this is kind of the same thing as just doing:

``` python
plotter.screenshot(filename)
```

except now we do something like
``` python
img = plotter.screenshot(None, return_img=True)
img.save(filename)
```
(but with extra tweaks because img is a 2D array here, not ImageData, may need to use `Pillow`)

---

Should I revert this change then, and stick with using `plotter.screenshot` directly to save the file here?

MatthewFlamm: I think we should probably get this PR in whatever form makes sense and then simplify the figure saving logic separately.  It is a separate issue.  

> compare_images reads and compares the png files, so are they not using the same image?

`pyvista.compare_images` uses `image_from_window` when a Plotter is passed in.


https://github.com/pyvista/pyvista/blob/8b1b5896e6c01cb362cc5f4b3af879093022629f/pyvista/plotting/utilities/regression.py#L145-L150

This is different logic than using `pyvista.screenshot` in ways that are not clear to me.  It seems that the difference, if any is not large, but it raises lots of questions.

pytest_pyvista/pytest_pyvista.py
user27182: `image_cache_dir` and `generated_image_dir` both have a default set by the parser. This means these options always have a value and neither will ever be `None`.
https://github.com/pyvista/pytest-pyvista/blob/10952115b004451f5afa8382ff6dd3ccb022d103/pytest_pyvista/pytest_pyvista.py#L55-L59
https://github.com/pyvista/pytest-pyvista/blob/10952115b004451f5afa8382ff6dd3ccb022d103/pytest_pyvista/pytest_pyvista.py#L40-L44

pytest_pyvista/pytest_pyvista.py
user27182: > I think we should probably get this PR in whatever form makes sense and then simplify the figure saving logic separately. It is a separate issue.

I have reverted to using `plotter.screenshot` for now for saving the images and saving is independent of `generated_image_dir` [ffaf8fd](https://github.com/pyvista/pytest-pyvista/pull/171/commits/ffaf8fd3e5a87d7394edef8f6f62e14c46df9dae)

> > compare_images reads and compares the png files, so are they not using the same image?
> 
> `pyvista.compare_images` uses `image_from_window` when a Plotter is passed in.
> 
> https://github.com/pyvista/pyvista/blob/8b1b5896e6c01cb362cc5f4b3af879093022629f/pyvista/plotting/utilities/regression.py#L145-L150
> 
> This is different logic than using `pyvista.screenshot` in ways that are not clear to me. It seems that the difference, if any is not large, but it raises lots of questions.

This does seem odd, and agree it should probably be looked into and fixed (separately).

MatthewFlamm:
I can re-go through this one, but it seems like it will interact with #175 logic.  Do you have a preference for which should go in first?

user27182:
Yea probably will be a merge conflict, maybe go with this one first since this one is ready but the other one looks like it needs some tweaks still

pytest_pyvista/pytest_pyvista.py
```diff
@@ -198,21 +223,27 @@ def __call__(self, plotter):  # noqa: ANN001, ANN204, C901, PLR0912
...
 
         if ((self.add_missing_images and not os.path.isfile(image_filename)) or self.reset_image_cache) and not self.reset_only_failed:  # noqa: PTH113
             plotter.screenshot(image_filename)
 
         if self.generated_image_dir is not None:
-            gen_image_filename = os.path.join(self.generated_image_dir, test_name[5:] + ".png")  # noqa: PTH118
+            gen_image_filename = os.path.join(self.generated_image_dir, image_name)  # noqa: PTH118
             plotter.screenshot(gen_image_filename)
 
+        if self.failed_image_dir is not None and not Path(image_filename).is_file():
```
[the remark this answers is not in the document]
user27182: Correct, the current logic:
``` python
if not os.path.isfile(image_filename) and self.fail_extra_image_cache and not self.reset_image_cache:
```
does not guarantee that a `RegressionFileNotFound` error will be raised, and hence it's possible for `pv.compare_images` to fail due to `FileNotFound` error.

But yes we should probably move this up so that, under the specific conditions above (where a `RegressionFileNotFound` _will_ be raised, the generated image will be saved first before the error is raised.

user27182:
> I think this one should probably go in first? Then, afterwards PRs like #177?

Sounds reasonable to me

PR #173, opened 2025-06-16
Test downstream tests with `pytest-xdist`

user27182:


akaszynski:
Thanks for fixing the merge.

akaszynski:
Compared with:
https://github.com/pyvista/pytest-pyvista/actions/runs/16125715470/job/45502024651

I'm seeing around a 3 minute improvement in test times:
https://github.com/pyvista/pytest-pyvista/actions/runs/16127303206/job/45507174163?pr=173

**Specifics:**
5m 15s for this branch for `Unit Testing` step in `Downstream tests` vs 8m 29s without `pytest-xdist`.

user27182:
Yes we saw substantial reduction in times downstream too in https://github.com/pyvista/pyvista/pull/7624. For the limited tests in this repo, we seem to not have any issues with the parallelism. But the full CI suite downstream was seeing concurrency issues across `test_plotting` and `test_renderer`, and we appear to have race conditions with reading mesh files with all the fixtures, predominantly with `core` testing... which led to some refactoring of fixtures but currently hit a snag with https://github.com/pyvista/pyvista/pull/7631 since test coverage is not reporting properly with all the fixtures...

user27182:
It's also a bit of an unknown if features like `disallow_unused_cache` from https://github.com/pyvista/pytest-pyvista/pull/169 will work without a chance of breaking, since that option uses a single global variable to store test states... not sure how that's managed in parallel.

MatthewFlamm:
pytest-mpl has this workflow in a PR, although it hasn't been implemented yet.  It would be useful to keep an eye on the progress there.

https://github.com/matplotlib/pytest-mpl/pull/242

PR #96, opened 2023-10-12
Add generated_image_dir ini configuration

bjlittle:
This pull-request adds the convenience of specifying the `generated_image_dir` as a `pytest` INI configuration option, complimenting its associated CLI option.

This is akin to the support provided for the `image_cache_dir` option.

PR #174, opened 2025-06-21
Set default generated image dir to `None`

user27182:
The docs state no images are saved unless this option is set:
https://github.com/pyvista/pytest-pyvista/blob/10952115b004451f5afa8382ff6dd3ccb022d103/pytest_pyvista/pytest_pyvista.py#L100-L102

Though, this is for `VerifyImageCache.__init__` (not sure if this class is used this way without the fixture / pytest options). The README is less clear about this:
https://github.com/pyvista/pytest-pyvista/blob/10952115b004451f5afa8382ff6dd3ccb022d103/README.rst?plain=1#L107-L108

But it seems to me that the intention of this option was for it to originally be optional. When this option was initially added in https://github.com/pyvista/pytest-pyvista/pull/22, it was not enabled by default

But, when the `pytest.ini` option was added in https://github.com/pyvista/pytest-pyvista/pull/96, this added a default value, which means this dir is never `None`, and hence images are _always_ saved.
https://github.com/pyvista/pytest-pyvista/blob/10952115b004451f5afa8382ff6dd3ccb022d103/pytest_pyvista/pytest_pyvista.py#L40-L44

This PR reverts this change and makes generating the directory optional once again.

### DIFF
--- a/pytest_pyvista/pytest_pyvista.py
+++ b/pytest_pyvista/pytest_pyvista.py
@@ -39,7 +39,7 @@ def pytest_addoption(parser) -> None:  # noqa: ANN001
     )
     parser.addini(
         "generated_image_dir",
-        default="generated_image_dir",
+        default=None,
         help="Path to dump test images from the current run.",
     )
     group.addoption(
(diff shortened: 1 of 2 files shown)


MatthewFlamm (approved):
I think it is a good idea to have this one off by default as the generated images could be large.  In the vast majority of cases only the failed images need to be checked.  With #171, this will be possible

bjlittle (approved):
Agreed. This is a sensible default :+1:

PR #175, opened 2025-06-22
Add `allow_unused_generated` option as a replacement for `fail_extra_image_cache`

### LINKED ISSUE #36: Conflicting flags do not raise an error or warning.
The flags `--fail_extra_image_cache` and `--add_missing_images` are incompatible with each other. This is not documented anywhere. We should:

- Update the documentation to cover this.
- Choose what flag will take priority in case of conflict. 
- Raise a warning in the code to notify that both flags are incompatible when both flags are active, and tell the user which one is prioritized.

user27182:
Follow-up to https://github.com/pyvista/pytest-pyvista/pull/92. Instead of simply renaming the argument, this PR also changes the default behavior so that an error is now raised by default if a generated image has no use.

I think this also resolves #36. The name `--allow_unused_generated` doesn't conflict with `--add_missing_images`, since enabling `--add_missing_images` is considered to be a usage of the images. This is documented in the README.

### DIFF
--- a/pytest_pyvista/pytest_pyvista.py
+++ b/pytest_pyvista/pytest_pyvista.py
@@ -199,7 +199,7 @@ class VerifyImageCache:
         image_filename = os.path.join(self.cache_dir, image_name)  # noqa: PTH118
 
         overwrite_cache = (self.add_missing_images or self.reset_image_cache) and not self.reset_only_failed
-        if not Path(image_filename).isfiles() or overwrite_cache:
+        if not Path(image_filename).is_file() or overwrite_cache:
             if self.allow_unused_generated or overwrite_cache:
                 plotter.screenshot(image_filename)
             else:
--- a/pytest_pyvista/pytest_pyvista.py
+++ b/pytest_pyvista/pytest_pyvista.py
@@ -221,13 +221,13 @@ class VerifyImageCache:
         # "test_" to get the name for the image.
         image_name = test_name[5:] + ".png"
         image_filename = os.path.join(self.cache_dir, image_name)  # noqa: PTH118
-        gen_image_filename = os.path.join(self.generated_image_dir, image_name)  # noqa: PTH118
+        gen_image_filename = None if self.generated_image_dir is None else os.path.join(self.generated_image_dir, image_name)  # noqa: PTH118
 
         if not os.path.isfile(image_filename) and not (self.allow_unused_generated or self.add_missing_images or self.reset_image_cache):  # noqa: PTH113
             # Raise error since the cached image does not exist and will not be added later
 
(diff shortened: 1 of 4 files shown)


pytest_pyvista/pytest_pyvista.py
```diff
@@ -197,14 +198,15 @@ def __call__(self, plotter):  # noqa: ANN001, ANN204, C901, PLR0912
...
-            # Make sure this doesn't get called again if this plotter doesn't close properly
-            plotter._before_close_callback = None  # noqa: SLF001
-            msg = f"{image_filename} does not exist in image cache"
-            raise RegressionFileNotFound(msg)
-
-        if ((self.add_missing_images and not os.path.isfile(image_filename)) or self.reset_image_cache) and not self.reset_only_failed:  # noqa: PTH113
-            plotter.screenshot(image_filename)
+        overwrite_cache = (self.add_missing_images or self.reset_image_cache) and not self.reset_only_failed
+        if not Path(image_filename).is_file() or overwrite_cache:
+            if self.allow_unused_generated or overwrite_cache:
```
MatthewFlamm, line 203: I personally find the kludging of these two type of uses together in a set of nested if blocks to be confusing to reason with in my head.  Maybe related to below as well.

This logic means that if `self.allow_unused_generated` is `True` and there is no image cache file,  then it adds the image cache file?  Shouldn't it be required to also be using `self.add_missing_images = True` or `self.reset_only_failed` in this case?

MatthewFlamm: What is optimal in understanding is definitely subjective.  I find the separate usage more understandable in that the first blocks checks if there should be an error raised.  The second block only checks if an image should be generated.  The logic then follows from those goals.  When combined, there are competing objectives in each branch for which outcome you are trying to understand.

MatthewFlamm:
This looks good to me but I don't see the specific test here that uses the new `--allow-unused-generated`` option.  Maybe doing #176 and #177 first would help, but let me know your preference in order here.

user27182:
> This looks good to me but I don't see the specific test here that uses the new `--allow-unused-generated`` option. Maybe doing #176 and #177 first would help, but let me know your preference in order here.

This case wasn't handled properly, so it's a good thing tests were added in this PR with [a1d10a7](https://github.com/pyvista/pytest-pyvista/pull/175/commits/a1d10a7891422f8907483457653c5cfd7e3c6d47) and [4aa4297](https://github.com/pyvista/pytest-pyvista/pull/175/commits/4aa42974e5e6ddb4d67ad5d6c95f5f96cace9a33)

If this one still looks good to you, we can merge this first, then #177, then #176

PR #176, opened 2025-06-24
Improve test coverage

user27182:
Use `--branch` to report partial misses and add tests for branches without coverage.

### DIFF
--- a/pytest_pyvista/pytest_pyvista.py
+++ b/pytest_pyvista/pytest_pyvista.py
@@ -197,20 +197,14 @@ class VerifyImageCache:
         image_name = test_name[5:] + ".png"
         image_filename = os.path.join(self.cache_dir, image_name)  # noqa: PTH118
 
-        if not os.path.isfile(image_filename):  # noqa: SIM102, PTH113
-            if self.fail_extra_image_cache:  # noqa: SIM102
-                if not self.reset_image_cache:
-                    # Make sure this doesn't get called again if this plotter doesn't close properly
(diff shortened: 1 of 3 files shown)


user27182:
Coverage is reporting misses... Will need to investigate

pytest_pyvista/pytest_pyvista.py
```diff
@@ -248,11 +248,6 @@ def remove_plotter_close_callback() -> None:
             # comparison to avoid a FileNotFoundError
             return
 
-        if self.failed_image_dir is not None and not Path(image_filename).is_file():
-            # Image comparison will fail, so save image before error
-            self._save_failed_test_images("error", plotter, image_name)
-            remove_plotter_close_callback()
```
user27182, line 254: This branch was identified as possibly redundant and unnecessary in https://github.com/pyvista/pytest-pyvista/pull/171#discussion_r2164247004, and https://github.com/pyvista/pytest-pyvista/pull/175 fixed the earlier logic such that, with branch coverage enabled by this PR, this branch has been identified as missing coverage and hence is not executed and is therefore not needed.

user27182:
Coverage report is working, and now only a single (partial) miss is reported. The missing branch is for the `windows_skip_image_cache` flag, which I was not able to add a patch test for under Linux due to a `pathlib` error trying to use `WindowsPath` under Linux

akaszynski (approved):
I see no issues with this, recommending merge for a release this weekend.

PR #187, opened 2025-07-23
Add `image_name_prefix` option

### LINKED ISSUE #186: Dynamically set the validation image directory
We ought to allow for a mechanism to automatically create a directory called `validation_images` adjacent to the current module being tested. This would help better organize the image cache in large projects

I'm currently working on an internal project with 100s (approaching 1000) validation images. This is spread across ~20 different sub-packages each with their own `tests/` directory. Rather than having one global validation image directory, I'd like to organize images so that they are adjacent to the module being tested.

I think we could configure the pytest-pyvista fixture to automatically determine that path based on the current module and set up that directory.

My ideal structure would be:

```txt
package/
├── module_a/
│   ├── ...
│   └── tests/
│       ├── validation_images/
│       │   └── image_a.png
│       └── test_module_a.py
├── module_b/
│   ├── ...
│   └── tests/
│       ├── validation_images/
│       │   └── image_b.png
│       └── test_module_b.py
└── ...
```

user27182:
Resolve #186

Instead of dynamically setting a validation directory though, the package name and/or module name is prepended to the test image name, see https://github.com/pyvista/pytest-pyvista/issues/186#issuecomment-3105599112

### DIFF
--- a/pytest_pyvista/pytest_pyvista.py
+++ b/pytest_pyvista/pytest_pyvista.py
@@ -7,17 +7,17 @@ from pathlib import Path
 import platform
 import shutil
 from typing import TYPE_CHECKING
+from typing import Callable
 from typing import Literal
 from typing import cast
 import warnings
 
 import pytest
 import pyvista
+from pyvista import Plotter
 
 if TYPE_CHECKING:  # pragma: no cover
     from collections.abc import Generator
-
-    from pyvista import Plotter
 
 
 VISITED_CACHED_IMAGE_NAMES: set[str] = set()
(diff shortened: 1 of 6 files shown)


user27182:
@banesullivan I believe the ``--image_name_prefix package`` option is most applicable to your use case from https://github.com/pyvista/pytest-pyvista/issues/186, or maybe even ``--image_name_prefix package+module``.

For `pyvista`, I think we should use ``--image_name_prefix module`` to make the test source explicit. This is done in https://github.com/pyvista/pyvista/pull/7744

pytest_pyvista/pytest_pyvista.py
```diff
@@ -328,24 +348,60 @@ def _make_failed_test_image_dir(
...
+    base = test_name.removeprefix("test_")
+    name_parts = []
+
+    if image_name_prefix not in {None, "none"} and node_path is not None:
+        if image_name_prefix in {"package", "full", "package+module"}:
+            # Find the `tests` dir and append its parent's name as the package
+            for parent in node_path.parents:
+                if parent.name == "tests":
+                    package_root = parent.parent
+                    package_name = package_root.name
```
MatthewFlamm, line 362: Does this mean that the package name is just the parent folder name?  What about a project that has a structure like scipy that can have a structure like: 

```txt
package
  subpackage1
    tests
      test_module.py
    subfolder
      tests
        test_module1.py
```

user27182: It's not intended to reflect the actual package name the root, ie scipy in this case. But really just to distinguish images from different test directories like in #186. Maybe `package` isn't the right term to use

user27182: The `full` option also used this assumption, but maybe we can try to identify the actual project root, and give the full path relative to that instead. This would probably be the best option to use in this case for the given structure to make the path explicit (and maybe also a bit verbose)

MatthewFlamm: Maybe this is what the code already does and it just needs different documentation?

MatthewFlamm: Thanks, I think I have my head around this better.  I suspect that we mostly need to change the naming here.  I will read it again.

PR #188, opened 2025-07-28
Fix cache directory not being absolute

edabor:
When temporary changing the working directory while calling `show` on a plotter instance, the image cache dir is not found since it is not relative to the `pytest` rootpath.
This PR fixes it.

```python
import pyvista as pv

pv.OFF_SCREEN = True
import contextlib
from pathlib import Path

def test_imcache(verify_image_cache, tmp_path: Path):
    sphere = pv.Sphere()
    plotter = pv.Plotter()
    plotter.add_mesh(sphere, color="red")
    with contextlib.chdir(tmp_path):
        plotter.show()
```
```python
pytest_pyvista.pytest_pyvista.RegressionFileNotFoundError: image_cache_dir\imcache.png does not exist in image cache
```

### DIFF
--- a/tests/test_pyvista.py
+++ b/tests/test_pyvista.py
@@ -759,7 +759,7 @@ def test_disallow_unused_cache_name_mismatch(testdir, disallow_unused_cache) -> 
         assert result.ret == pytest.ExitCode.OK
 
 
-@pytest.mark.skipif(sys.version_info < (3, 11))
+@pytest.mark.skipif(sys.version_info < (3, 11), reason="Needs contextlib.chdir")
 def test_cache_generated_dir_relative(testdir: pytest.Testdir) -> None:
     """
     Test that directories (cache and generated) are relative to test root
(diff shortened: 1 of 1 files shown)


pytest_pyvista/pytest_pyvista.py
```diff
@@ -418,11 +418,16 @@ def verify_image_cache(request, pytestconfig) -> Generator[VerifyImageCache, Non
...
 
-    cache_dir = _get_option_from_config_or_ini(pytestconfig, "image_cache_dir")
+    cache_dir = Path(_get_option_from_config_or_ini(pytestconfig, "image_cache_dir"))
     gen_dir = _get_option_from_config_or_ini(pytestconfig, "generated_image_dir")
     failed_dir = _get_option_from_config_or_ini(pytestconfig, "failed_image_dir")
 
-    verify_image_cache = VerifyImageCache(request.node.name, cache_dir, generated_image_dir=gen_dir, failed_image_dir=failed_dir)
+    verify_image_cache = VerifyImageCache(
+        request.node.name,
+        pytestconfig.rootpath / cache_dir,
```
user27182, line 427: Does this return a `Path` object? This keyword is currently typed as `str`, so should be updated accordingly

beroda: Done in 28f54b5c4b5d0271af2cba970c74d84ea515eb65

user27182 (approved):

Issue #172, opened 2025-06-16
`verify_image_cache` not called when setting `auto_close=False`


user27182:
The `test_ssao_pass` test:

https://github.com/pyvista/pyvista/blob/5f8fea51169091ce25f83db99e29387f816b4e29/tests/plotting/test_plotting.py#L3118-L3129
``` python
def test_ssao_pass():
    ugrid = pv.ImageData(dimensions=(2, 2, 2)).to_tetrahedra(5).explode()
    pl = pv.Plotter()
    pl.add_mesh(ugrid)

    pl.enable_ssao()
    pl.show(auto_close=False)

    # ensure this fails when ssao disabled
    pl.disable_ssao()
    with pytest.raises(RuntimeError):
        pl.show()
```

apparently currently passes CI, since no errors are currently raised by pytest on `main`. But, if the test is passing, this also means that with #171, the `ssao_pass` image should _not_ be copied into the `failed_image_dir`. But, this _is_ being copied into the failed images directory. I tested this locally and this line definitely executes:
https://github.com/pyvista/pytest-pyvista/blob/10952115b004451f5afa8382ff6dd3ccb022d103/pytest_pyvista/pytest_pyvista.py#L227

But `pytest` isn't failing the test for some reason. When I make this change though:
``` diff
-    pl.show(auto_close=False)
+    pl.show()
```
Then pytest properly raises an error and fails the test. So the `auto_close` keyword is somehow silencing this error. Maybe there is a bare Except somewhere catching the `RegressionError` and preventing it from being raised?

user27182:
I think this error partly stems from #189, or at least it seems that this PR exacerbates this issue. If #189 is reverted and `--disallow_unused_cache` is enabled, I suspect CI will pass. This is tested in #199

user27182:
This test passed before #189 
``` py
import pyvista as pv

def test_callback(verify_image_cache):
    verify_image_cache.skip = True
    pl = pv.Plotter(off_screen=True)
    pl.show(auto_close=False)
    assert pl._before_close_callback is not None
    pl.close()
```

But the assertion fails with #189. Hence, there is no callback, and `verify_image_cache.__call__` isn't called, and hence the test doesn't get added to  `VISITED_CACHED_IMAGE_NAMES` set.

beroda:
It seems that the `weakref` is messing things up. When removing it, the above test works.
@akaszynski what was the motivation of using a weak ref here ? I'm not an expert on the topic.

[...]

PR #189, opened 2025-07-29
Fix callback

beroda:
The `verify_image_cache` object is not called when a `before_close_callback` is already injected in `Plotter.show()`.

```python
import pyvista as pv

pv.OFF_SCREEN = True

def test_imcache(verify_image_cache, mocker):
    sphere = pv.Sphere()
    plotter = pv.Plotter()
    plotter.add_mesh(sphere, color="red")
    plotter.show(before_close_callback=(m := mocker.MagicMock()))

    m.assert_called_once_with(plotter)
```
```python
E               Failed: Fixture `verify_image_cache` is used but no images were generated.
E               Did you forget to call `show` or `plot`, or set `verify_image_cache.allow_useless_fixture=True`?.
```

This PR fixes it by fully wrapping the call to `Plotter.show`, injecting the `verify_image_cache` object on-the-fly after the user callback.
It induces that `pv.global_theme.before_close_callback` property is not modified.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -41,7 +41,7 @@ tests = [
   "numpy<2.3",
   "pytest-cov==6.2.1",
   "pytest>=3.5.0",
-  "pytest_mock",
+  "pytest_mock<3.15",
 ]
 
 [project.entry-points."pytest11"]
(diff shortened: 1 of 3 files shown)


user27182:
This PR caused a breakage with the `--disallow_unused_cache` option, see #199
I think some images are somehow not being added to the `VISITED_CACHED_IMAGE_NAMES` set. I suspect this has something to do with tests that use `auto_close=False`, see #172

PR #201, opened 2025-08-30
Fix auto_close preventing image cache call

edabor:
Closes https://github.com/pyvista/pytest-pyvista/issues/172

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -108,8 +108,17 @@ jobs:
           which python
           pip list
 
+      # run unit testing with both parallel and without parallel flags but
+      # using `--parallel-mode` flag to ensure coverage files are unique, see
+      # https://coverage.readthedocs.io/en/latest/cmd.html
       - name: Unit Testing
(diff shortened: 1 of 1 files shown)


akaszynski:
> remove the @pytest.mark.usefixtures('no_images_to_verify') mark and add a pl.close() to make sure the plot is closed by the test and not the check_gc fixture

Option 2 seems more explicit. Let's go with that.

tests/test_pyvista.py
```diff
@@ -855,3 +855,29 @@ def test_imcache(verify_image_cache, tmp_path: Path, pytestconfig: pytest.Config
...
+            pl.close()
+
+        def test_auto_close_2(verify_image_cache, mocker):
+            verify_image_cache.skip = True
+            plotter = pv.Plotter()
+            plotter.show(auto_close=False, before_close_callback=(m:=mocker.MagicMock()))
+            plotter.close()
+            m.assert_called_once_with(plotter)
+        """
+    )
```
akaszynski, line 880: Nitpick: I'd prefer this be a PY_FILE at the top of the test (or somewhere nearby) ease readability.

akaszynski (approved):
LGTM aside from a nitpick.

PR #198, opened 2025-08-27
Add `--doc_mode` config option for testing documentation images

user27182:
Add `--doc_mode` for testing documentation images. This effectively moves most of the documentation tests from `pyvista` to `pytest-pyvista`. 

Code is adapted from https://github.com/pyvista/pyvista/blob/main/tests/doc/tst_doc_build.py

The functionally of the tests is largely the same as the current PyVista tests _except_ for how flaky tests are handled. Instead of having a separate flaky_tests dir, subdirectories inside the cache image dir itself are allowed, and you just place the multiple images there instead

PR #211, opened 2025-09-02
Allow using multiple cache images

### LINKED ISSUE #197: Add a `flaky_tests` directory
The PyVista docs tests have a `flaky_tests` dir (see https://dev.pyvista.org/extras/developer_notes.html#documentation-image-regression-testing). This is useful because it allows specifying more than one "valid" image for the test. A similar feature already in `pytest-pyvista` is to flag a test as `high_variance` to allow for the additional error for different images.

But, using `high_variance` has drawbacks because if a test has really high variance (e.g. above the threshold of 1000), the test is usually completely skipped, either based on OS or based on VTK version, which leaves gaps in the test coverage. Using  a `flaky_tests` dir instead would allow for more flexibility since tests with really high variance can just use multiple images instead of skipping it altogether. A good use case for this is with the recent updates to `vtkCubeAxesActor` which broke many of our tests (see https://github.com/pyvista/pyvista/pull/7682). Due to this change, the variance was too high for VTK < 9.6 compared to the latest dev wheels, so the tests were completely skipped for lower VTK versions. If we instead had a `flaky_tests` dir, we could keep coverage for both older and new VTK by storing multiple "correct" versions of the test images.

user27182:
Resolve #197 

Instead of a separate `flaky_tests` dir (as suggested in #197), this PR allows subdirectories inside the cache dir directly. To specify multiple images, just place them inside a sub-dir with the test name. This is consistent with the changes in #198.

This is also tested downstream in https://github.com/pyvista/pyvista/pull/7914

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -153,7 +153,7 @@ These are the flags you can use when calling ``pytest`` in the command line:
   directory, relative to `pytest root path <https://docs.pytest.org/en/latest/reference/reference.html#pytest.Config.rootpath>`.
   This will override any configuration, see below.
 
-* ``--generate_subdirs <NAME>`` saves generated test images in separate sub-directories
+* ``--generate_subdirs`` saves generated test images in separate sub-directories
   instead of saving them directly to the ``generated_image_dir``. Without this option,
   generated images are saved as ``generated_image_dir/<test_name>.png``; with this
   option enabled, they are instead saved as


README.rst
```diff
@@ -120,6 +150,14 @@ These are the flags you can use when calling ``pytest`` in the command line:
   directory, relative to `pytest root path <https://docs.pytest.org/en/latest/reference/reference.html#pytest.Config.rootpath>`.
   This will override any configuration, see below.
 
+* ``--generate_subdirs <NAME>`` saves generated test images in separate sub-directories
+  instead of saving them directly to the ``generated_image_dir``. Without this option,
+  generated images are saved as ``generated_image_dir/<test_name>.png``; with this
+  option enabled, they are instead saved as
+  ``<generated_image_dir>/<test_name>/<image_name>.png``, where the image name has
+  the format `<system>_<python-version>_<pyvista-version>_<vtk-version>`. This can
```
akaszynski, line 158:
```suggestion
  the format ``<system>_<python-version>_<pyvista-version>_<vtk-version>``. This can
```

akaszynski (approved):
One nitpick, otherwise LGTM.

beroda:
> This was sort of previously supported as a `--generated_image_name` arg that could be unspecified (use default env info) or provide a fully custom input. It was changed / renamed / removed in [eb49e26](https://github.com/pyvista/pytest-pyvista/commit/eb49e26e0dcb69a5b40228cb8abab0ee90fdac79)
> 
> Is this the kind of customization you would prefer? I removed it because (1) the arg name wasn't great, since passing the flag `generated_image_name` _without_ providing a name seemed confusing/ambiguous about what it meant, and (2) actually providing this info is challenging in CI (How to get vtk version and pass it as an arg?).
> 
> Can you provide more details about what kind of API you want for customizing this? We can add a _separate_ flag for appending a prefix or suffix? Something like `--subdir_image_name_suffix`?

Agreed that it can be tough to specify the prefix via command-line, my request was not clear sorry.
I'm 100% for leaving `--generate_subdirs` as a `store_true`.

 I was more thinking of having an attribute to `VerifyImageCache` to enable such kind of customization:

```python
@pytest.fixture(autouse=True)
def verify_image_cache_wrapper(verify_image_cache):
    import vtk, mylib
    verify_image_cache.subdir_image_prefix = f"{vtk.__version__}_{mylib.__version__}"
    return verify_image_cache
```
It would default to `ENV_INFO` if not provided.

README.rst
```diff
@@ -105,6 +105,39 @@ If you need to use any flag inside the tests, you can modify the
...
+.. code-block:: bash
+
+    cache/my_image/0.jpg
+    cache/my_image/1.jpg
+
+.. note::
+
+   - The sub-directory name should match the name of the test.
+   - The image names in sub-directories can be arbitrary, e.g. ``0.jpg`` or
+   ``foo.jpg``.
```
beroda, line 131:
```suggestion
     ``foo.jpg``.
```

user27182:
Allowing customization in the fixture directly makes sense. 

I am planning on adding more options for env info (e.g. check for GitHub RUNNER_ENVIRONMENT to append 'self-hosted', or include rendering info like EGL or GPU manufacturer). Maybe we can add these along with your request in a follow-up PR?

I am planning on adding `--generate_subdirs` option to `--doc_mode` as well from #198. How can we generalize your feature request to allow customization the name for doc mode as well, since we can't customize a fixture for that case?

Maybe we can try to address all of this and harmonize / extend the API after this PR and the doc mode PR merge?

beroda (approved):

PR #214, opened 2025-09-04
Allow customizing env info

user27182:
Allow customizing info saved in image filenames when using `--generate_subdirs`. See https://github.com/pyvista/pytest-pyvista/pull/211#pullrequestreview-3183786750 for details.

Adds a new dataclass with the option to customize it using `verify_image_cache.env_info`.

### DIFF
--- a/pytest_pyvista/pytest_pyvista.py
+++ b/pytest_pyvista/pytest_pyvista.py
@@ -29,6 +29,11 @@ if TYPE_CHECKING:  # pragma: no cover
 VISITED_CACHED_IMAGE_NAMES: set[str] = set()
 SKIPPED_CACHED_IMAGE_NAMES: set[str] = set()
 
+try:
+    _GPU_VENDOR = pyvista.GPUInfo().vendor
+except Exception:  # noqa: BLE001 # pragma: no cover
+    _GPU_VENDOR = "UNKNOWN"
+
 
 @dataclass
 class _EnvInfo:
@@ -37,6 +42,7 @@ class _EnvInfo:
     python: bool = True
     pyvista: bool = True
     vtk: bool = True
+    gpu: bool = True
     runner: bool = True
     suffix: str = ""
 
@@ -45,8 +51,19 @@ class _EnvInfo:
         python_version = f"py-{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}" if self.python else ""
         pyvista_version = f"pyvista-{pyvista.__version__}" if self.pyvista else ""
         vtk_version = f"vtk-{vtkmodules.__version__}" if self.vtk else ""
-        runner = os.environ.get("RUNNER_ENVIRONMENT", "local") if self.runner else ""
-        values = [f"{self.prefix}", f"{system_version}", f"{python_version}", f"{pyvista_version}", f"{vtk_version}", f"{runner}", f"{self.suffix}"]
+        gpu = f"gpu-{_GPU_VENDOR}" if self.gpu else ""
+        runner = os.environ.get("RUNNER_ENVIRONMENT", "local-hosted") if self.runner else ""
+
+        values = [
+            f"{self.prefix}",
+            f"{system_version}",
+            f"{python_version}",
+            f"{pyvista_version}",
+            f"{vtk_version}",
+            f"{gpu}",
+            f"{runner}",
+            f"{self.suffix}",
+        ]
         return "_".join(val for val in values if val)
 
     @staticmethod
--- a/pytest_pyvista/pytest_pyvista.py
+++ b/pytest_pyvista/pytest_pyvista.py
@@ -70,14 +70,14 @@ class _EnvInfo:
         return "macOS" if system == "Darwin" else system
 
     @staticmethod
-    def _gpu_vendor() -> str:  # pragma: no cover
+    def _gpu_vendor() -> str:
         # Get cached value
         if _GPU_VENDOR[0]:
             return _GPU_VENDOR[0]
 
         try:
             vendor = pyvista.GPUInfo().vendor
-        except Exception:  # noqa: BLE001
+        except Exception:  # noqa: BLE001  # pragma: no cover
             vendor = "UNKNOWN"
 
         # Try to shorten vendor string
--- a/pytest_pyvista/pytest_pyvista.py
+++ b/pytest_pyvista/pytest_pyvista.py
@@ -42,18 +42,18 @@ class _EnvInfo:
     pyvista: bool = True
     vtk: bool = True
     gpu: bool = True
-    host: bool = True
+    ci: bool = True
     suffix: str = ""
 
     def __repr__(self) -> str:
         os_info = _EnvInfo._get_os()
         os_version = f"{os_info[0]}-{os_info[1]}" if self.os else ""
         machine = f"{platform.machine()}" if self.machine else ""
+        gpu = f"gpu-{_EnvInfo._gpu_vendor()}" if self.gpu else ""
         python_version = f"py-{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}" if self.python else ""
         pyvista_version = f"pyvista-{pyvista.__version__}" if self.pyvista else ""
         vtk_version = f"vtk-{vtkmodules.__version__}" if self.vtk else ""
-        gpu = f"gpu-{_EnvInfo._gpu_vendor()}" if self.gpu else ""
-        host = os.environ.get("RUNNER_ENVIRONMENT", "local-hosted") if self.host else ""
+        ci = f"{'' if os.environ.get('CI', None) else 'no'}-CI" if self.ci else ""
 
         values = [
             f"{self.prefix}",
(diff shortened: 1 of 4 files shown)


akaszynski (approved):
This is a great way to add metadata to our images just through the file name. We might also consider adding metadata to the jpeg itself, perhaps containing much of the `pyvista.Report()`, but having it there in the file name might be enough.

user27182:
> > This is a great way to add metadata to our images just through the file name. We might also consider adding metadata to the jpeg itself, perhaps containing much of the `pyvista.Report()`, but having it there in the file name might be enough.
> 
> Didn't know that was possible. Will try adding that.

Will leave this for a separate PR, and after #213 is implemented, because the unit tests save PNGs, but the doc tests save JPGs, so storing the metadata will need to consider both formats.

pytest_pyvista/pytest_pyvista.py
```diff
@@ -28,23 +30,90 @@
...
+            return name, version
+        name = "macOS" if system == "Darwin" else system
+        return name, platform.release()
 
-ENV_INFO = _get_env_info()
+    @staticmethod
+    def _gpu_vendor() -> str:
+        # Get cached value
+        if _GPU_VENDOR[0]:
+            return _GPU_VENDOR[0]
```
beroda, line 89: Using a cached property seems appropriate here
```suggestion
    from functools import cached_property

    @cached_property
    def _gpu_vendor(self) -> str:
```

pytest_pyvista/pytest_pyvista.py
```diff
@@ -28,23 +30,90 @@
...
+            text = vendor  # pragma: no cover
+        # Shorten original string and remove whitespace
+        vendor = vendor[: len(text)].replace(" ", "")
+        # Remove all potentially invalid/undesired filename characters
+        disallowed = r'[\\/:*?"<>|\s.\x00]'
+        vendor = re.sub(disallowed, "", vendor)
+
+        # Cache the value globally
+        _GPU_VENDOR[0] = vendor
+        return _GPU_VENDOR[0]
```
beroda, line 116: With `cached_property` suggestion
```suggestion
        return vendor
```

pytest_pyvista/pytest_pyvista.py
```diff
@@ -28,23 +30,90 @@
...
+class _EnvInfo:
+    prefix: str = ""
+    os: bool = True
+    machine: bool = True
+    python: bool = True
+    pyvista: bool = True
+    vtk: bool = True
+    gpu: bool = True
+    ci: bool = True
+    suffix: str = ""
```
beroda, line 46: If a user wants to override everything, he/she needs to set `False` to all attributes of the instance.
Maybe we could add a `custom` attribute to simplify this.

```suggestion
    suffix: str = ""
    custom: str = ""
```

pytest_pyvista/pytest_pyvista.py
```diff
@@ -28,23 +30,90 @@
...
+            f"{pyvista_version}",
+            f"{vtk_version}",
+            f"{ci}",
+            f"{self.suffix}",
         ]
-    )
+        return "_".join(val for val in values if val)
 
+    @staticmethod
+    def _get_os() -> tuple[str, str]:
```
beroda, line 72: Maybe a `cached_property` can also be appropriate ?
```suggestion
    @cached_property
    def _get_os(self) -> tuple[str, str]:
```

pytest_pyvista/pytest_pyvista.py
```diff
@@ -28,23 +30,90 @@
...
+    @staticmethod
+    def _gpu_vendor() -> str:
+        # Get cached value
+        if _GPU_VENDOR[0]:
+            return _GPU_VENDOR[0]
+
+        try:
+            vendor = pyvista.GPUInfo().vendor
+        except Exception:  # noqa: BLE001  # pragma: no cover
+            vendor = "UNKNOWN"
```
beroda, line 94:
```suggestion
            return "unknown"
```

tests/test_pyvista.py
```diff
@@ -1013,3 +1025,70 @@ def test_imcache(verify_image_cache):
...
             assert file_has_changed(str(from_test), str(from_cache))
+
+
+def test_env_info() -> None:
+    """Test env info dataclass."""
+    info = str(_EnvInfo())
+    os_info = _EnvInfo._get_os()  # noqa:SLF001
+    assert " " not in info
+    assert info.startswith(os_info[0] + "-" + os_info[1])
+    if platform.system() == "Linux" and sys.version_info >= (3, 10):
```
beroda, line 1036: I think we should mock calls to `platform.system` (and GPU vendor, os ...) to improve coverage and make test platform-independent.

I can help with that if needed.

pytest_pyvista/pytest_pyvista.py
```diff
@@ -223,6 +292,7 @@ class VerifyImageCache:
     add_missing_images = False
     reset_only_failed = False
     generate_subdirs = None
+    env_info: _EnvInfo
```
beroda, line 295: The attribute should be set in `VerifyImageCache` instance rather than for the class since it can be different for each test.
```suggestion
```

pytest_pyvista/pytest_pyvista.py
```diff
@@ -598,6 +670,7 @@ def verify_image_cache(
     VerifyImageCache.add_missing_images = pytestconfig.getoption("add_missing_images")
     VerifyImageCache.reset_only_failed = pytestconfig.getoption("reset_only_failed")
     VerifyImageCache.generate_subdirs = pytestconfig.getoption("generate_subdirs")
+    VerifyImageCache.env_info = _EnvInfo()
```
beroda, line 673: See my suggestion to set the attribute on the instance rather than on the class.
```suggestion
```

beroda (changes requested):
Thanks for the responsive implementation

I added a few suggestions, happy to contribute if not being clear.

pytest_pyvista/pytest_pyvista.py
user27182: We could also allow `EnvInfo | str` for this use case. And just set it directly to the desired string and bypass the dataclass entirely.

tests/test_pyvista.py
user27182: Ok. We should maybe consider adding dedicated macOS and Windows runners.

The try...except for Linux was only needed to support python 3.9 too. So if that's dropped (in line PyVista downstream), we can simplify this code a bit too.

pytest_pyvista/pytest_pyvista.py
user27182: A new EnvInfo instance _is_ created with each test already though, so I guess this is more a style thing to store the value during the VerifyImageCache init instead of immediately after init?

pytest_pyvista/pytest_pyvista.py
user27182: Agreed, this is not where it should be set. Do you prefer it be included as a new init kwarg? Or just set it directly after init.

pytest_pyvista/pytest_pyvista.py
user27182: This was already possible, (`str(self.env_info)` is used), but tests/docs are added in [79d6f59](https://github.com/pyvista/pytest-pyvista/pull/214/commits/79d6f59a8a3ece39dc7adc6c0a828d7673d9d200)

pytest_pyvista/pytest_pyvista.py
user27182: EnvInfo is initialized on the instance in [c5c8b49](https://github.com/pyvista/pytest-pyvista/pull/214/commits/c5c8b497796256a23eb8441b1d5d2d700ef70a51)

pytest_pyvista/pytest_pyvista.py
```diff
@@ -238,6 +314,7 @@ def __init__(  # noqa: PLR0913
     ) -> None:
         """Initialize VerifyImageCache."""
         self.test_name = test_name
+        self.env_info: str | _EnvInfo = _EnvInfo()
```
user27182, line 317: Maybe call this attribute `generated_image_name` instead? Since that's what this is really setting? Can make it a formal property too and add validation to raise an error if setting this without also using `--generate_dirs`

beroda (approved):
The `env_info` name could be improved https://github.com/pyvista/pytest-pyvista/pull/214#discussion_r2328289421 but overall LGTM.

pytest_pyvista/pytest_pyvista.py
user27182: I guess it's a trade off, since using `env_info` makes it clear that the dataclass with env info is being modified, but using `subdirs_image_name` makes it clear that the image name is being modified. So for maximum clarify maybe two separate properties are needed? 

I think this is fine for now, setting a custom string is the non-default option for power users, and is documented in the docs example, so having this use-case as the slightly confusing one in terms of attribute names is probably Ok.

beroda: Agreed, no best solution

PR #216, opened 2025-09-07
Add config option to save test images as JPG or PNG

user27182:


tkoyama010 (approved):
LGTM

user27182:
> > There's a mysterious error with the tests using `jpg` format and/or when using cache images in nested directories that popped up after merging #214... still trying to resolve this. Can't seem to reproduce it locally and so is proving hard to debug.
> 
> I think this may be CI-specific, since running all the tests locally on macOS in isolation works. Skipping some jpg format tests with nested dirs OR if vtk < 9.3 resolves this. Both these cases are niche use-cases, and even then I think in practice this will still work since I'm thinking it's a CI issue, so I'm calling it good.

This error is fixed by closing all plotters between tests with `pyvista.close_all()`

PR #224, opened 2025-09-10
Validate the image cache files

user27182:
This PR:

- Ensures that jpg files are not used when the image format is png, and vice-versa
- Ensures that test names are unique, i.e. check that a subdir and image file do not share the same name


akaszynski (approved):
Good idea to add a check beforehand. This is good to go.

PR #225, opened 2025-09-10
Validate doc mode CLI args are only used with `--doc_mode`

user27182:
Raise an error if using any doc-specific args without also using `--doc_mode`.

### DIFF
--- a/pytest_pyvista/pytest_pyvista.py
+++ b/pytest_pyvista/pytest_pyvista.py
@@ -39,6 +39,10 @@ SKIPPED_CACHED_IMAGE_NAMES: set[str] = set()
 DEFAULT_ERROR_THRESHOLD: float = 500.0
 DEFAULT_WARNING_THRESHOLD: float = 200.0
 
+
+_PLUGIN_PARSER_GROUP = "pyvista"
+_PLUGIN_PARSER: pytest.Parser  # global reference to the parser
+
 _ImageFormats = Literal["png", "jpg"]
 
 
@@ -152,9 +156,12 @@ class RegressionFileNotFoundError(RegressionFileNotFound):
 
 def pytest_addoption(parser) -> None:  # noqa: ANN001
     """Adds new flag options to the pyvista plugin."""  # noqa: D401
+    global _PLUGIN_PARSER  # noqa: PLW0603
+    _PLUGIN_PARSER = parser
+
     _add_common_pytest_options(parser)
 
-    group = parser.getgroup("pyvista")
+    group = parser.getgroup(_PLUGIN_PARSER_GROUP)
     group.addoption(
         "--reset_image_cache",
         action="store_true",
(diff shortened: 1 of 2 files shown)


akaszynski:
> But, I also think this kind of API is a bit clunky, and so perhaps modifying the API as suggested in https://github.com/pyvista/pytest-pyvista/pull/225#issuecomment-3277143912 instead is better. Open for discussion

I'm always a fan of:

> "There should be one-- and preferably only one --obvious way to do it"

Since the documentation test config has grown, we might consider having a section in `pyproject.toml` that allows us to specify a named config and place all the configuration there rather than having a slew of CLI args. We can always have simple args for a straightforward input and even consider adding in a `ini` file for those that wish to have the config out of `pyproject.toml`, but I'm seeing the community shift everything to the TOML.

user27182:
> > But, I also think this kind of API is a bit clunky, and so perhaps modifying the API as suggested in https://github.com/pyvista/pytest-pyvista/pull/225#issuecomment-3277143912 instead is better. Open for discussion
> 
> 
> 
> I'm always a fan of:
> 
> 
> 
> > "There should be one-- and preferably only one --obvious way to do it"

Indeed this would be best for CLI args since the two modes are mutually exclusive. It would be better if we could do:

- `pytest --image_format jpg` for unit tests
- `pytest --doc_mode --image_format jpg` for documentation tests

Rather than force users to use `--doc_image_format` (with a doc prefix) for the doc mode tests, which is what the current API requires.

 
> Since the documentation test config has grown, we might consider having a section in `pyproject.toml` that allows us to specify a named config and place all the configuration there rather than having a slew of CLI args.

All of the new doc mode options are available both as CLI args and config options. I think allowing users to specify options both ways is good. 

The issue is that users may want to specify _different_ options for doc mode than unit tests. This is why the `doc_` prefix is used, so we can specify

```
image_format = 'png'
doc_image_format = 'jpg'
```

So, if we _don't_ use the `doc_` prefix for CLI args, but then _require_ the prefix for the config file, then that sort of makes the API inconsistent (different arg name in CLi vs config). 


We could address this by accepting non-prefixed args in the config file by default for _both_ unit tests and doc mode tests, i.e. writing

```
image_format = 'jpg'
```

Would use jpg format for both use cases, equivalent to doing:

- `pytest --image_format jpg` for unit tests
- `pytest --doc_mode --image_format jpg` for documentation tests

And if users want to specify different options for the two use cases, we allow use of a `doc_` prefix in the config file to distinguish them, as mentioned earlier:

```
image_format = 'png'
doc_image_format = 'jpg'
```

So in general, the API for doc mode would use non-prefixed options by default, but allow using the prefix in config files for cases where users want different values for the two modes.

user27182:
> So in general, the API for doc mode would use non-prefixed options by default, but allow using the prefix in config files for cases where users want different values for the two modes.

I think this is probably the way to go. Once #226 is merged, the APIs for both doc mode and regular unit tests will have almost identical options for `image_format`, `generate_subdirs`, `image_cache_dir`, `generated_image_dir`, and `failed_image_dir`. We could drop the use of the `doc_` prefix, and update the README documentation to combine these into three sections:

1. a `Common global flags` section for these shared args
2. a `Unit tests global flags` section for options specific to the unit tests
3. a `Documentation tests global flags` section for options specific to doc mode

Then we could close this PR since using `--doc_` prefixes would no longer be allowed for CLI args

PR #226, opened 2025-09-11
Add hook for customizing doc mode test cases

user27182:
Follow-up to #213

Add a hook users can use to customize test cases, similar to how test cases for unit tests can be customized.

This PR extends #213 to specifically expose the `env_info` attribute. Future PRs could extend this further to customize the error or warning threshold, add skip options, etc, similar to what can be done with the current `verify_image_cache` fixture for unit tests.


akaszynski:
> This is ready, but also includes all the changes from #213, so should be reviewed after that one

Approved and merged #213. I'll leave the merge conflicts up to you for the moment.

PR #227, opened 2025-09-14
Add interactive file tests for vtksz files with doc mode

### LINKED ISSUE #220: Move `vtksz` file size tests to `pytest-pyvista`
#198 added _most_ of the documentation test code from `pyvista` to `pytest-pyvista`, but there are still tests for checking the vtksz file sizes which we should also move.

user27182:
Resolve #220 

- [x] Add tests for maximum vtksz file size
- [x] Convert vtksz files to screenshots and include them as part of generated test images
- [x] Add new ``--include_vtksz`` flag to enable these tests 
- [x] Make work with pytest-xdist
- [x] Update docs


akaszynski:
> This was tested upstream with [pyvista/pyvista#7878](https://github.com/pyvista/pyvista/pull/7878)
> It works, but it took approx 40 mins (without using parallel testing) to convert the 1540 `vtksz` files to screenshots using `playwright`:

Considering we've had a few bad interactive images, I see no issue with including this in our testing. Ideally we could get the time down to ~20 minutes using parallel testing. `pyvista-plot` directive should just render static images using `RenderWindow`, and I'm prefer to keep separation of concerns.

We have enough servers to support this testing, and I have two spares should this end up bogging down our CI.

user27182:
> > This was tested upstream with [pyvista/pyvista#7878](https://github.com/pyvista/pyvista/pull/7878)
> > It works, but it took approx 40 mins (without using parallel testing) to convert the 1540 `vtksz` files to screenshots using `playwright`:
> 
> Considering we've had a few bad interactive images, I see no issue with including this in our testing. Ideally we could get the time down to ~20 minutes using parallel testing. `pyvista-plot` directive should just render static images using `RenderWindow`, and I'm prefer to keep separation of concerns.
> 
> We have enough servers to support this testing, and I have two spares should this end up bogging down our CI.

Thanks for the feedback. Let's keep this feature as part of `pytest-pyvista` then. I'll see if I can reduce the test times. If you happen to be familiar with taking screenshots of html files and can think of some other way to speed this up let me know...

pytest_pyvista/doc_mode.py
```diff
@@ -66,56 +71,142 @@ def optional_dir_with_temp(option: str, prefix: str) -> Path:
...
+    with sync_playwright() as p:
+        browser = p.chromium.launch(headless=True)
+        page = browser.new_page(viewport={"width": DEFAULT_IMAGE_WIDTH, "height": DEFAULT_IMAGE_HEIGHT})
+        page.goto(f"file://{html_file}")
+        page.screenshot(path=output_path)
+        browser.close()
+    with Image.open(output_path) as im:
+        # Resize since the browser does not respect the requested size
+        im.resize((DEFAULT_IMAGE_WIDTH, DEFAULT_IMAGE_HEIGHT)).save(output_path)
+    return output_path
```
user27182, line 206: Setting `viewport` in `browser.new_page` does not seem to work to actually set the screen size. Seems that only the aspect ratio is modified, and the outputs are `1024x768` even though `400x300` is requested. I tried using `screen` keyword as well, and also using CLI args to set this when launching chromium, but none of this worked. Might be able to reduce test time spent generating the screenshots if we can figure this out.

user27182:
> > This was tested upstream with [pyvista/pyvista#7878](https://github.com/pyvista/pyvista/pull/7878)
> > It works, but it took approx 40 mins (without using parallel testing) to convert the 1540 `vtksz` files to screenshots using `playwright`:
> 
> Considering we've had a few bad interactive images, I see no issue with including this in our testing. Ideally we could get the time down to ~20 minutes using parallel testing. 
> 
> We have enough servers to support this testing, and I have two spares should this end up bogging down our CI.

Reduced to 17 mins by using a single browser context with [584da66](https://github.com/pyvista/pytest-pyvista/pull/227/commits/584da666caee0da16556da39360ab3491a9bba73), see the run here:

https://github.com/pyvista/pyvista/actions/runs/17775690805/job/50522678419?pr=7878#step:13:51840

And I tried using `-n8` for the tests to reduce this further and got worker errors, will need to make these tests thread-safe.

akaszynski:
> Reduced to 17 mins by using a single browser context with [584da66](https://github.com/pyvista/pytest-pyvista/pull/227/commits/584da666caee0da16556da39360ab3491a9bba73), see the run here:

17 minutes is more than good! Nice work.

user27182:
> > > This was tested upstream with [pyvista/pyvista#7878](https://github.com/pyvista/pyvista/pull/7878)
> > > It works, but it took approx 40 mins (without using parallel testing) to convert the 1540 `vtksz` files to screenshots using `playwright`
> 
> > Reduced to 17 mins by using a single browser context with [584da66](https://github.com/pyvista/pytest-pyvista/pull/227/commits/584da666caee0da16556da39360ab3491a9bba73)
> > And I tried using `-n8` for the tests to reduce this further and got worker errors, will need to make these tests thread-safe.
> 
> Reduced to ~10 mins by using `multiprocessing` with two processes: https://github.com/pyvista/pyvista/actions/runs/17784701827/job/50554089479?pr=7878

Now down to < 8 mins with `-n4` https://github.com/pyvista/pyvista/actions/runs/17818297584/job/50655695509?pr=7878#step:13:16489

pytest_pyvista/doc_mode.py
```diff
@@ -169,25 +343,55 @@ def add_to_dict(filepath: Path, key: str) -> None:
...
+
+        input_is_vtksz = input_path is not None and input_path.suffix == ".vtksz"
+        cache_is_from_vtksz = cache is not None and cache.stem.endswith("_vtksz")
+
+        # Interactive test cases from vtksz files are mutually exclusive with regular image file tests
+        if input_is_vtksz or cache_is_from_vtksz:
+            if vtksz:
+                test_cases_list.append(test_case)
+        elif not vtksz:
+            test_cases_list.append(test_case)
```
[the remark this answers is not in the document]
user27182: Fixed in [39a7e04](https://github.com/pyvista/pytest-pyvista/pull/227/commits/39a7e041689ec4daa7576731b024df971c6bcf39)

user27182:
Ready for review

pytest_pyvista/doc_mode.py
```diff
@@ -51,83 +82,237 @@ def require_existing_dir(option: str) -> Path:
...
+
+        with path.open("rb") as file:
+            data = file.read()
+        output_path = Path(output_dir) / f"{path.stem}.html"
         output_paths.append(output_path)
+        embed_data_to_viewer_file(data, output_path)
+    return output_paths
+
+
+def _html_screenshots(html_files: list[Path], output_dir: Path, *, log: bool = True) -> list[Path]:
```
user27182, line 249: Logging is currently enabled by default and cannot be configured. For the PyVista repo, this will print more than 3000 lines for the 1500+ images.

This should be configurable, but I would prefer to do this in a separate PR.

PR #230, opened 2025-09-18
Drop support for Python 3.9 and VTK 9.1

user27182:
Python 3.9 reaches EOL on October 30. I propose dropping support for it. `pyvista` has already done so downstream.

My main motivation is that I'm seeing seg faults with #227 for vtk 9.1 with python 3.9, and so some tests are skipped, but coverage is only reported for 3.9, and hence missing coverage is being reported even though it is covered by later versions.

I don't think these seg faults are necessarily inherent to vtk 9.1, and may be an issue with our CI (tests that were previously passing are now failing as additional tests were added), but instead of trying to isolate and fix this I think it may be better to drop support completely based on python 3.9 EOL.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -59,8 +59,6 @@ jobs:
 
       matrix:
         include:
-          - python-version: "3.9"
-            vtk-version: "9.1"
           - python-version: "3.10"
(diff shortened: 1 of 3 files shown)


akaszynski (approved):

tkoyama010 (approved):
LGTM

PR #232, opened 2025-09-21
Remove 'doc' prefix CLI args

user27182:
Update the API to remove the separation between doc mode args and unit test args. Common args between the two modes now use the same name. For doc mode, the order of priority is now:

1. Use the value from CLI (if arg is passed)
2. Use the value from the config with a `doc_` prefix (if not `None`)
3. Fallback to using the value from the config _without_ a `doc_` prefix.

Closes #225. More info: https://github.com/pyvista/pytest-pyvista/pull/225#issuecomment-3289097688


README.rst
```diff
@@ -296,51 +193,87 @@ The tests have three main modes of failure:
...
+* Use ``--allow_useless_fixture`` to prevent test failure when the ``verify_image_cache``
+  fixture is used but no images are generated. If no images are generated (i.e. there are
+  no calls made to ``Plotter.show()`` or ``mesh.plot()``), then these tests will fail
+  by default. Set this CLI flag to allow this globally, or use the test-specific flag
+  by the same name below to configure this on a per-test basis.
+
+Documentation testing flags
+===========================
+These flags are specific to documentation tests. They cannot be used with regular unit
+tests.
```
user27182, line 266: The term "cannot" is a bit misleading here. Technically they can be used, but they don't do anything. I can add CLI input validation (separate PR) to raise an error if an unsupported argument is used for the wrong test mode

README.rst
```diff
@@ -105,10 +105,10 @@ cache images.
...
 Unit tests
 ----------
-Once installed, you only need to use the command `pl.show()` in your test. The
+Once installed, you only need to use the command ``pl.show()`` in your test. The
 plugin will automatically manage the cache generation if it does not exist, and
-the image comparison itself. Make sure you enable `pv.OFF_SCREEN` when loading
-PyVista, so the `pl.show()` doesn't pop up any window while testing.  By default,
+the image comparison itself. Make sure you enable ``pv.OFF_SCREEN`` when loading
+PyVista, so the ``pl.show()`` doesn't pop up any window while testing.  By default,
 the verify_image_cache fixture should be used for each test for image comparison:
```
akaszynski, line 112:
```suggestion
the ``verify_image_cache`` fixture should be used for each test for image comparison:
```

README.rst
```diff
@@ -135,7 +135,7 @@ a wrapped version could be used
 
 
 If you need to use any flag inside the tests, you can modify the
-`verify_image_cache` object in the test
+``verify_image_cache`` object in the test
```
akaszynski, line 138:
```suggestion
``verify_image_cache`` object in the test:
```

README.rst
```diff
@@ -296,51 +193,87 @@ The tests have three main modes of failure:
...
-* ``--doc_images_dir <DIR>``> sets the target directory of images to be tested.
-
-  .. note::
+------------
+These are the flags you can use when calling ``pytest`` in the command line.
 
-     With Sphinx, build images are typically saved to ``doc/_build/html/_images``.
+Common flags
+============
+These flags may be used with regular unit testing or with ``--doc_mode`` enabled.
```
akaszynski, line 205:
```suggestion
These flags may be used with regular unit testing or documentation testing with
``--doc_mode`` enabled.
```

README.rst
akaszynski: That would be preferable. Silent failures aren't great; it would be preferable to instead raise a helpful error that indicates if the CLI arg isn't available for unit/doc testing. Let's add that in a followup PR.

pytest_pyvista/doc_mode.py
```diff
@@ -83,24 +84,35 @@ def require_existing_dir(option: str) -> Path:
...
-        cls.doc_image_format = cast("_AllowedImageFormats", _get_option_from_config_or_ini(config, "doc_image_format"))
-        cls.doc_generate_subdirs = bool(_get_option_from_config_or_ini(config, "doc_generate_subdirs"))
+        cls.image_format = cast("_AllowedImageFormats", _get_option_from_config_or_ini(config, "image_format"))
+        cls.generate_subdirs = bool(_get_option_from_config_or_ini(config, "generate_subdirs"))
 
         cls.include_vtksz = bool(_get_option_from_config_or_ini(config, "include_vtksz"))
 
+        if not any(cls.doc_images_dir.iterdir()) and not any(cls.image_cache_dir.iterdir()):
+            msg = (
+                "No doc images or cache images found! The doc images dir:\n"
```
akaszynski, line 108:
```suggestion
                "No doc images or cache images found. The doc images dir:\n"
```

README.rst
```diff
@@ -375,7 +308,59 @@ These are the flags you can use when calling ``pytest`` in the command line:
...
+* ``macos_skip_image_cache``: For test where the plotting in MacOS is different
+  from Windows/Linux.
+
+* ``skip``: If you have a test that plots a figure, but you don't want to compare
+  its output against the cache, you can skip it with this flag.
+
+* ``allow_useless_fixture``: Set this flag to ``True`` to prevent test failure when the
+  ``verify_image_cache`` fixture is used but no images are generated. The value of this
+  flag takes precedence over the global flag by the same name (see above).
+
```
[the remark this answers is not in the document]
user27182: It's still a valid option though:
https://github.com/pyvista/pytest-pyvista/blob/635a98a2c8f1f8751d44e03972f50ec58990441c/pytest_pyvista/pytest_pyvista.py#L202-L206

Do we want to keep the option but just remove it from the documentation?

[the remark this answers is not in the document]
akaszynski: Correct. It's already a shared flag, no reason to repeat it.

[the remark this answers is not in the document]
akaszynski: I get it now. Understood. Let's keep it.

PR #236, opened 2025-09-23
Refactor `pytest_addoption` common options

user27182:
Currently, `_add_common_cli_and_ini_options` is called twice (with `doc=True`, and `doc=False`). But, the only thing this does is append a second INI option with a `doc_` prefix. This PR adds `_add_common_ini_option` to do this instead so that `_add_common_cli_and_ini_options` only needs to be called once.

To better separate the addition of unit test options from doc options, these are also nested into separate functions.


akaszynski:
Think we're ready for a release @user27182?

user27182:
> Think we're ready for a release @user27182?

Yes. Maybe include #237 too?

PR #244, opened 2025-10-05
Register placeholder `xdist` hook if `xdist` not installed

### LINKED ISSUE #243: v0.3.0 broken without `pytest-xdist` installed
The `pytest_configure_node` hook is specific to `pytest-xdist`. If not installed, we get this error:
https://github.com/pyvista/pyvista/actions/runs/18263354544/job/51995182324#step:6:1

user27182:
Fix #243


user27182:
@akaszynski I think we'll need a `0.3.1` patch with this fix

akaszynski (approved):

PR #263, opened 2026-01-06
Minimum python version not fullfilled

edabor:
I was able to install the latest `pytest-pyvista` version with `python < 3.10`, even though `3.10` should be the minimum version since `v0.3.0` thanks to https://github.com/pyvista/pytest-pyvista/pull/230.

image

This PR fixes it.
See related pyproject specification https://peps.python.org/pep-0621/#requires-python

image

Should we backport that fix to `0.3.0`, `0.3.1` and `v0.3.2` ? This can impact users since it raises TypeError for `python 3.9`:
```
    _OriginalImageFormats = _AllowedImageFormats | Literal["gif", "vtksz"]
TypeError: unsupported operand type(s) for |: '_LiteralGenericAlias' and '_LiteralGenericAlias'
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,8 +20,8 @@ dynamic = ["description", "version"]
 license = "MIT"
 license-files = ["LICENSE"]
 name = "pytest_pyvista"
-python_requires = ">=3.10"
 readme = "README.rst"
+requires-python = ">=3.10"
 
 [project.urls]
 Home = "https://github.com/pyvista/pytest-pyvista"


edabor:
> This is similar to how if you try to install PyVista with python 3.14, which it can't due to vtk not being supported, the resolver tries installing an older PyVista instead when the vtk version was not pinned, so it tries installing pyvista v0.37 or something like that.

Wow, indeed it installs `pyvista==0.36` ... without `vtk` and without complaints.

image

akaszynski:
> Should we backport that fix to 0.3.0, 0.3.1 and v0.3.2 ? This can impact users since it raises TypeError for python 3.9:

We can't really backport unless we yank those old versions. If we're fine with yanking, I'm fine with it.

akaszynski:
> If we're fine with yanking, I'm fine with it.

@user27182, your thoughts?

PR #267, opened 2026-03-03
feat: avoid reading the cache when using --ignore_image_cache argument

germa89:
Currently, the cache is read even if we are ignoring it using `--ignore_image_cache`.

This is unnecessary and generates performance penalties.

I started work on this while I tried to find a way to disable the plugin for an specific pytest run (for debugging purposes). The closer you can get to this behaviour is using `--ignore_image_cache`.

Alternatively, we could go for a more explicit flag like `--disable-image-check`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -205,16 +205,16 @@ Common flags
 These flags may be used with regular unit testing or documentation testing with
 ``--doc_mode`` enabled.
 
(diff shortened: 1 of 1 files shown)


germa89:
@user27182 I implemented the requested changes.

akaszynski (approved):
Thanks @germa89. @user27182, calling this good to go, but feel free to comment if I missed something.

PR #196, opened 2025-08-13
test parallel in downstream tests

akaszynski:
Check if downstream tests can be run in parallel with `pytest-xdist`.

PR #271, opened 2026-03-23
Use `tox` for unit testing

edabor:
This PR introduces `tox` as CI handler in order to simplify CI configuration and enhance reproducibility with local testing.


edabor:
> To be fair the codecov comment gave misleading checkmarks of approval ✅. I think project settings are the problem for not enforcing this and making coverage thresholds a requirement.
> 
> We lost coverage before too: [#196 (comment)](https://github.com/pyvista/pytest-pyvista/pull/196#issuecomment-3239601310)
> 
> Which I believe was fixed by #202. Might be helpful for fixing the issue here too.

I'll refactor coverage settings in another PR if it is fine for you.

PR #278, opened 2026-03-26
Add VTK resource cleanup and macOS autorelease pool fixtures

### LINKED ISSUE #272: VTK Resource Cleanup Fixtures
Every project testing with PyVista needs to close plotters and force garbage collection between tests. Without this, VTK render windows accumulate, exhaust X11 client connections (Linux), exhaust Cocoa autorelease pools (macOS), or leave dangling OpenGL contexts that cause segfaults under pytest-xdist.

Right now, every downstream project I've worked on has had to independently writes:

```python
@pytest.fixture(autouse=True)
def _close_plotters():
    yield
    pv.close_all()
```

I found this exact fixture (with minor variations) in four of my (closed source) projects, Four copies, same code, but each forgot to add `gc.collect()` because I forgot I needed it until segfaults started randomly occuring...

### What PyVista does

PyVista's `tests/plotting/conftest.py` calls `pv.close_all()` inside the `check_gc` fixture, then `gc.collect()` to force VTK C++ destructors to fire. Without the explicit GC pass, render windows and their OpenGL contexts linger. This is invisible in serial test runs but causes collisions across xdist workers sharing GPU resources.

### Proposal

Add an autouse, function-scoped fixture to pytest-pyvista:

```python
@pytest.fixture(autouse=True)
def _pyvista_cleanup():
    yield
    pv.close_all()
    gc.collect()
```

Always active. No configuration needed. Projects that already call `pv.close_all()` in their own conftest see no change (double-closing is safe). Projects that don't get cleanup for free.

Add a config option to disable if someone has a reason to:

```toml
[tool.pytest.ini_options]
pyvista_auto_cleanup = true  # default: true
```

### Related issues

- [pyvista#3869](https://github.com/pyvista/pyvista/issues/3869) Segfault creating/closing many Plotters in a loop
- [pyvista#2509](https://github.com/pyvista/pyvista/issues/2509) Memory grows across doctests without `close_all`/`gc.collect`
- [pyvista#1248](https://github.com/pyvista/pyvista/issues/1248) Hanging VTK references after plotter close

### LINKED ISSUE #273: macOS Apple Silicon Autorelease Pool
VTK's `vtkCocoaRenderWindow` creates Objective-C objects in an `NSAutoreleasePool`. Without draining between tests, these accumulate. On Apple Silicon, this caps out around 500 render windows per process, then SIGBUS.

Under pytest-xdist this is worse. Each worker is a bare Python process with no main run loop to drain the pool automatically. Tests that create multiple render windows (common in multi-view or isolation tests) hit the limit fast.

I ran into this debugging segfaults in my closed project under pytest-xdist. The primary cause turned out to be an application-level VTK reference leak, but the autorelease pool fixture is defense-in-depth that prevents a separate accumulation problem. PyVista already has this fixture. Downstream projects shouldn't need to rediscover it.

### What PyVista does

`tests/plotting/conftest.py`:

```python
@pytest.fixture(autouse=True)
def macos_memory_leak(request):
    from Foundation import NSAutoreleasePool
    pool = NSAutoreleasePool.alloc().init()
    yield
    del pool
```

Conditioned on `platform.system() == 'Darwin' and platform.machine() == 'arm64'`. No-op everywhere else. Gracefully skips if `pyobjc` isn't installed.

### Proposal

Add this as an autouse fixture in pytest-pyvista. No-op on non-macOS, non-ARM64, and when `Foundation` isn't available. Zero impact on Linux/Windows CI.

```python
@pytest.fixture(autouse=True)
def _pyvista_macos_autorelease():
    if sys.platform != 'darwin' or platform.machine() != 'arm64':
        yield
        return
    try:
        from Foundation import NSAutoreleasePool
    except ImportError:
        yield
        return
    pool = NSAutoreleasePool.alloc().init()
    yield
    del pool
```

This one has no reason to be opt-out. It either applies or doesn't.

This has been undocumented tribal knowledge living only in PyVista's `tests/plotting/conftest.py`.

banesullivan:
Closes #272, closes #273

- Add `_pyvista_close_plotters` autouse fixture that calls `pv.close_all()` + `gc.collect()` after each test. This prevents VTK render window accumulation, X11 connection exhaustion (Linux), Cocoa autorelease pool overflow (macOS), and dangling OpenGL context segfaults under pytest-xdist.
- Add `pyvista_close_all` INI option (default: `true`) to disable the cleanup fixture for projects that need custom cleanup.
- Add `_pyvista_macos_autorelease` autouse fixture that drains the Cocoa `NSAutoreleasePool` between tests. VTK's `vtkCocoaRenderWindow` creates Objective-C objects that accumulate without draining, capping out around 500 render windows per process before SIGBUS. No-op on non-macOS, non-ARM64, and when `pyobjc` is not installed.
- Remove the hand-rolled `close_all_plotters` fixture from the test conftest since the plugin now provides it.

### DIFF
--- a/tests/test_pyvista.py
+++ b/tests/test_pyvista.py
@@ -1436,7 +1436,7 @@ def test_macos_autorelease_fixture_registered(pytester: pytest.Pytester) -> None
 
         def test_plotter_creates_and_closes():
             # This test just verifies the fixture doesn't interfere.
-            # On macOS ARM64 with pyobjc, it drains the autorelease pool.
+            # On macOS ARM64 it drains the autorelease pool.
             # Everywhere else, it's a no-op.
             pl = pv.Plotter()
             pl.add_mesh(pv.Sphere())
--- a/tests/test_pyvista.py
+++ b/tests/test_pyvista.py
@@ -1396,6 +1396,24 @@ def test_close_all_runs_by_default(pytester: pytest.Pytester) -> None:
     result.assert_outcomes(passed=2)
 
 
+def test_clear_trame_servers(pytester: pytest.Pytester) -> None:
+    """Test the trame servers registry is cleared after each test."""
+    pytester.makepyfile(
+        """
+        from trame.app.core import AVAILABLE_SERVERS, Server
+        from trame.app import get_server
+
+        def test_get_server():
+            assert isinstance(get_server("name"), Server)
+
+        def test_get_server_cleared():
+            assert len(AVAILABLE_SERVERS) == 0
+        """
+    )
+    result = pytester.runpytest("-v")
+    result.assert_outcomes(passed=2)
+
+
 def test_close_all_can_be_disabled(pytester: pytest.Pytester) -> None:
     """Setting pyvista_close_all = false should skip cleanup."""
     pytester.makeini(
--- a/tests/test_pyvista.py
+++ b/tests/test_pyvista.py
@@ -1370,3 +1370,57 @@ def test_max_image_size(pytester: pytest.Pytester, doc_mode, original_size, max_
     generated_image = pv.read(gen_file)
     expected_size = _get_thumbnail_size(original_size, max_image_size)
     assert generated_image.dimensions == (*expected_size, 1)
+
+
+def test_close_all_runs_by_default(pytester: pytest.Pytester) -> None:
+    """The _pyvista_close_plotters fixture should close plotters after each test."""
+    pytester.makepyfile(
+        """
+        import pyvista as pv
+        from pyvista.plotting.plotter import _ALL_PLOTTERS
+
+        pv.OFF_SCREEN = True
+
+        def test_plotter_is_open():
+            pl = pv.Plotter()
+            pl.add_mesh(pv.Sphere())
+            assert len(_ALL_PLOTTERS) > 0
+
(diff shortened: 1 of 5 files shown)


tests/test_pyvista.py
```diff
@@ -1370,3 +1370,95 @@ def test_{test_name}(verify_image_cache):
...
+    from Foundation import NSAutoreleasePool  # noqa: PLC0415
+
+    pool = NSAutoreleasePool.alloc().init()
+    # Create a plotter to generate Cocoa objects
+    pl = pv.Plotter(off_screen=True)
+    pl.add_mesh(pv.Sphere())
+    pl.render()
+    pl.close()
+    # Draining the pool should not crash
+    del pool
```
user27182: We currently don't have CI coverage for this (i.e it's always skipped). But on `macOS` this will always fail since `Foundation` is never installed. I recommend adding a macOS runner to test this.

user27182 (commented):
Recommend enabling macOS autorelease by default and adding macOS runner to ensure it's working.

tests/test_pyvista.py
banesullivan: Good call, added a MacOS job in bab5aa8

banesullivan:
Looks like CI is now passing, @edabor is this ready to ship now?

PR #279, opened 2026-03-30
Add codecov settings

edabor:
Add `codecov` settings (copy-paste from `pyvista`)

> To be fair the codecov comment gave misleading checkmarks of approval ✅. I think project settings are the problem for not enforcing this and making coverage thresholds a requirement.
> 
> We lost coverage before too: https://github.com/pyvista/pytest-pyvista/pull/196#issuecomment-3239601310
> 
> Which I believe was fixed by https://github.com/pyvista/pytest-pyvista/pull/202. Might be helpful for fixing the issue here too.
> 
> _Originally posted by @user27182 in https://github.com/pyvista/pytest-pyvista/issues/271#issuecomment-4136237998_

### DIFF
--- a/codecov.yaml
+++ b/codecov.yaml
@@ -1,0 +1,22 @@
+comment:
+  layout: "diff"
+  behavior: default
+
+coverage:
+  status:
+    project:
+      default:
+        # basic
+        target: 90%
+        threshold: 0%
+        # advanced
+        if_not_found: success
+        if_ci_failed: error
+        if_no_uploads: error
+    patch:
+      default:
+        # basic
+        target: 90%
+        if_not_found: success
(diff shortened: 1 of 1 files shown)


banesullivan (approved):

banesullivan:
> Instead of duplicating `codecov` settings all over the place, we could also set a global yaml for all repos in the `pyvista` organization [here](https://app.codecov.io/account/github/pyvista/yaml)

This is neat! However, I'd prefer our configuration as code be checked into a reposiotry somewhere as opposed to on codecov's web platform. Otherwise I will 100% forget where it is configured and be very confused when it automatically applies to some other repo in the org

PR #293, opened 2026-05-18
Disable PyVista SMP tools during test sessions for determinism

### LINKED ISSUE #288: Disable SMP tools during testing
With the addition of `enable_smp_tools` in pyvista, we should add a fixture like this to disable it automatically.

```py
# conftest.py                                                                                                                                  
import pytest                                                                                                                                  
import pyvista as pv                      
                                                                                                                                               
@pytest.fixture(scope='session', autouse=True)                     
def _disable_smp_in_tests():              
    pv.enable_smp_tools(backend='sequential', n_threads=1) 
```

banesullivan:
Parallel SMP can reorder floating-point reductions and perturb image regression results between runs. This adds a session-scoped autouse fixture that pins PyVista to the sequential single-thread SMP backend so renders are reproducible.

Default on, opt out with `pyvista_disable_smp = false`. Graceful no-op on versions of PyVista that predate `enable_smp_tools`.

resolves #288

### DIFF
--- a/pytest_pyvista/pytest_pyvista.py
+++ b/pytest_pyvista/pytest_pyvista.py
@@ -342,6 +342,13 @@ def pytest_addoption(parser: pytest.Parser) -> None:  # noqa: PLR0915
         help="Automatically close all plotters and run gc.collect() after each test (default: True).",
     )
 
+    parser.addini(
+        "pyvista_disable_smp",
+        type="bool",
+        default=True,
+        help=("Disable PyVista SMP (shared-memory parallel) tools during the test session for deterministic, reproducible runs (default: True)."),
+    )
+
 
 class VerifyImageCache:
     """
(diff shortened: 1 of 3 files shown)


pytest_pyvista/pytest_pyvista.py
```diff
@@ -342,6 +342,13 @@ def _add_doc_cli_and_ini_options() -> None:
         help="Automatically close all plotters and run gc.collect() after each test (default: True).",
     )
 
+    parser.addini(
+        "pyvista_disable_smp",
+        type="bool",
+        default=True,
+        help=("Disable PyVista SMP (shared-memory parallel) tools during the test session for deterministic, reproducible runs (default: True)."),
+    )
+
```
user27182, line 352: With all these recent PRs adding new config options, I am concerned about the clear separation between unit tests and doc mode tests. In code, these were previously separated with functions:

```
    _add_common_cli_and_ini_options()
    _add_unit_test_cli_and_ini_options()
    _add_doc_cli_and_ini_options()
```

I suggest moving these options into the appropriate function. Or if you prefer to _only_ add ini options, then please make a new function to better organize them.

If the option isn't relevant to `doc_mode`, we'll need to ensure the option does nothing or raises an error if trying to use it via CLI, etc.

PR #295, opened 2026-05-18
Compare best-match cached baseline first for multi-cache tests

### LINKED ISSUE #286: multi-cache selection bug
We're seeing warnings in PyVista's CI even when a correct cached image already exists.

This happens when a test has multiple cached images and the plugin compares against the first sorted image instead of the best match for the current environment. If that first image fails but a later image passes, the test is converted into a warning.

That warning is noisy and I'm pretty sure it is just because the wrong baseline was checked first.

`_get_file_paths()` returns a sorted list of cached images in a directory

https://github.com/pyvista/pytest-pyvista/blob/144eafb90b4a46c060412d0c497fe4cc0a9d18cd/pytest_pyvista/pytest_pyvista.py#L641-L643

`VerifyImageCache.__call__()` picks `cached_image_paths[0]` as the initial comparison target

https://github.com/pyvista/pytest-pyvista/blob/144eafb90b4a46c060412d0c497fe4cc0a9d18cd/pytest_pyvista/pytest_pyvista.py#L496

only after that first comparison fails does it try the remaining cached images and if a later cached image passes, the failure is downgraded into a warning

When a test intentionally keeps multiple valid baselines, for example by OS or VTK version, the plugin should try the best match first.

Right now it can produce `errors_as_warnings` or `warnings` for tests that are already correctly cached. In that case the warning is reporting cache ordering, not image instability.

### OS-specific cache example

[`tests/plotting/image_cache/camera3d_widget/`](https://github.com/pyvista/pyvista/tree/main/tests/plotting/image_cache/camera3d_widget) contains:

- `linux-vtk-9.6.png`
- `macOS-vtk-9.6.png`
- `windows-vtk-9.6.png`

On macOS, the files are sorted with `linux-vtk-9.6.png` first. `pytest-pyvista` compares against that image first, fails, and then succeeds against `macOS-vtk-9.6.png`.

Observed warning:

```text
camera3d_widget Exceeded image regression error of 500.0 with an image error equal to: 989.819607843139
This test has multiple cached images. It initially failed (as above) but passed when compared to:
    .../tests/plotting/image_cache/camera3d_widget/macOS-vtk-9.6.png
```

### VTK-specific cache example

[`tests/plotting/image_cache/plot_show_grid/`](https://github.com/pyvista/pyvista/tree/main/tests/plotting/image_cache/plot_show_grid) contains:

- `vtk-9.5.png`
- `vtk-9.6.png`

On VTK 9.6, the same pattern applies. The 9.5 image is checked first, then the 9.6 image passes.

This affects several PyVista tests with multi-cache baselines, including:

- `camera3d_widget`
- `plot_show_grid`
- `plot_show_grid_font_size[...]`
- `plot_show_bounds_params[...]`
- `contour_labels_smoothing_constraint[...]`
- `hide_cells[...]`
- `plot_label_fmt`

---

This inflates the new CI image report with repeated warnings that should be quiet.

It makes the report harder to review because legitimate image drift gets mixed together with warnings caused only by baseline selection order.

When multiple cached images exist, `pytest-pyvista` should compare against the best candidate first.

Suggested priority:

1. exact environment match, if encoded
2. exact OS + VTK match
3. exact VTK match
4. remaining cached images as fallback

At minimum, the plugin should avoid warning when an exact matching cached image already exists and passes.

This is what I was able to roughly prototype as a solution I'm just not sure if it should be in pytest-pyvista or pyvista directly since it is somewhat specific to pyvista's CI and regression image naming conventions

```py
# pytest_pyvista.py

import re

_IMAGE_CACHE_OS_ALIASES = {
    'darwin': ('darwin', 'macos'),
    'linux': ('linux',),
    'windows': ('windows',),
}
_KNOWN_IMAGE_CACHE_OS_TOKENS = frozenset(
    token for tokens in _IMAGE_CACHE_OS_ALIASES.values() for token in tokens
)


def _parse_cached_vtk_version(path: Path) -> tuple[int, int] | None:
    match = re.search(r'vtk-(\d+)\.(\d+)', path.stem.lower())
    if match is None:
        return None
    return int(match.group(1)), int(match.group(2))


def _cached_image_sort_key(path: Path, env_info: _EnvInfo) -> tuple[int, int, int, int, str]:
    stem = path.stem.lower()
    env_str = str(env_info).lower()

    if stem == env_str:
        env_rank = 0
    elif env_str in stem:
        env_rank = 1
    else:
        env_rank = 2

    system = platform.system().lower()
    current_os_tokens = _IMAGE_CACHE_OS_ALIASES.get(system, (system,))
    has_known_os_token = any(token in stem for token in _KNOWN_IMAGE_CACHE_OS_TOKENS)
    matches_current_os = any(token in stem for token in current_os_tokens)
    os_rank = 0 if matches_current_os else 1 if not has_known_os_token else 2

    current_vtk_version = tuple(pyvista.vtk_version_info[:2])
    cached_vtk_version = _parse_cached_vtk_version(path)
    if cached_vtk_version is None:
        vtk_rank = 1
        vtk_distance = 0
    else:
        vtk_rank = 0 if cached_vtk_version == current_vtk_version else 2
        vtk_distance = abs(cached_vtk_version[0] - current_vtk_version[0]) * 100 + abs(
            cached_vtk_version[1] - current_vtk_version[1]
        )

    return env_rank, os_rank, vtk_rank, vtk_distance, path.as_posix()


def _get_file_paths(dir_: Path, ext: str, *, env_info: _EnvInfo | None = None) -> list[Path]:
    """Get all paths of files with a specific extension inside a directory tree."""
    paths = sorted(dir_.rglob(f'*.{ext}'))
    if env_info is None:
        return paths
    return sorted(paths, key=lambda path: _cached_image_sort_key(path, env_info))


# In VerifyImageCache.__call__(), change this:
# cached_image_paths = _get_file_paths(image_dirname, ext=self.image_format) if image_dirname.is_dir() else [image_filename]
# to this:
cached_image_paths = (
    _get_file_paths(image_dirname, ext=self.image_format, env_info=self.env_info)
    if image_dirname.is_dir()
    else [image_filename]
)
```

banesullivan:
A test with multiple cached baselines (one per OS or VTK version) compared against the lexicographically first image. On a platform whose baseline sorted later, the first comparison failed and the test passed against the correct image but was downgraded to a noisy warning, even though the cache was correct. PyVista CI was full of these.

Cached candidates are now ranked by environment match (exact env string, current OS with darwin/macos aliasing, then exact and nearest VTK version) before comparison, so the right baseline is checked first and a correctly cached test stays silent. Token matching is boundary-aware so names like `non-linux-foo` or `myvtk-1.2` do not false-match.

Behavior note: for an existing multi-image cache the first compared image changes. It still falls back to every candidate, so pass/fail outcomes are unchanged; only the comparison order and the spurious warnings differ. Single-image caches and the string/None paths keep the previous lexicographic order.

resolves #286

### DIFF
--- a/tests/test_pyvista.py
+++ b/tests/test_pyvista.py
@@ -22,8 +22,11 @@ from pytest_pyvista.pytest_pyvista import _DOC_MODE_CLI_ARGS
 from pytest_pyvista.pytest_pyvista import _SYSTEM_PROPERTIES
 from pytest_pyvista.pytest_pyvista import _UNIT_TEST_CLI_ARGS
 from pytest_pyvista.pytest_pyvista import PARSER_GROUP_NAME
+from pytest_pyvista.pytest_pyvista import _cached_image_sort_key
 from pytest_pyvista.pytest_pyvista import _EnvInfo
+from pytest_pyvista.pytest_pyvista import _get_file_paths
 from pytest_pyvista.pytest_pyvista import _get_thumbnail_size
+from pytest_pyvista.pytest_pyvista import _parse_cached_vtk_version
 from pytest_pyvista.pytest_pyvista import _SystemProperties
 from pytest_pyvista.pytest_pyvista import pytest_addoption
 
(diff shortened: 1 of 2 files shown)


tests/test_pyvista.py
```diff
@@ -1061,6 +1064,147 @@ def test_imcache(verify_image_cache):
...
+    correct = Path(f"{os_token}-vtk-{vtk_major}.{vtk_minor}.png")
+    wrong_os = Path(f"otheros-vtk-{vtk_major}.{vtk_minor}.png")
+    no_tokens = Path("totally-unrelated-name.png")
+    paths = [wrong_os, no_tokens, correct]
+
+    ordered = sorted(paths, key=lambda p: _cached_image_sort_key(p, env_info))
+    assert ordered[0] == correct
+    # An unparsable name must not raise and must sort last.
+    assert _cached_image_sort_key(no_tokens, env_info)  # does not raise
+    assert ordered[-1] == no_tokens
```
user27182, line 1130: Highly recommend using `str(_EnvInfo()` here somewhere (or in a separate test), since this is how actual files are named with the ``--generate_subdirs`` option enabled. So we should make sure that the naming conventions used by that option work with the new sorting. Otherwise, are we relying on some sort of assumption for how users should otherwise format these file names, if they're not using `--generate_subdirs` to name them? Should we document that file names with specific tokens are sorted by priority when there are multiple cached images?


E.g.
``` py
correct = Path(f"{env_info}.png")
```
Which looks something like
```
macOS-23.4.0_arm64_gpu-Apple_py-3.14.0_pyvista-0.49.dev0_vtk-9.6.1_no-CI.png
```

user27182:
Superseded by #304

PR #297, opened 2026-05-18
Add opt-in blank rendered image detection

### LINKED ISSUE #238: Detect empty or mostly empty images
Add option to check for mostly blank images. E.g. compare image to a blank image (`Plotter().show()`), and raise error if error < threshold.

banesullivan:
A test that renders nothing still passes image regression when a matching blank baseline exists, which silently hides a broken test. This adds an optional check that compares the render against an empty plotter of the same window size and errors when the difference is within tolerance.

Off by default. Enable with `pyvista_check_blank_images = true`, tune sensitivity with `pyvista_blank_image_atol`, and bypass a specific test with `verify_image_cache.allow_blank_image = True`. A legitimately sparse scene (a small object on a large uniform background) can trip the check, which is why it is opt-in and the tolerance is configurable.

resolves #238

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -344,6 +344,12 @@ in the beginning of your test function.
   ``verify_image_cache`` fixture is used but no images are generated. The value of this
   flag takes precedence over the global flag by the same name (see above).
 
+* ``allow_blank_image``: When the global ``pyvista_check_blank_images`` option is enabled
+  (see `Configuration`_), every test that renders an essentially blank (near-uniform)
+  image fails, since this usually indicates a broken test that renders nothing yet still
+  passes. Set this flag to ``True`` to bypass the check for a test that intentionally
+  renders a near-blank scene.
+
 * ``env_info``: Dataclass for controlling the environment info used to name the generated
   test image(s) when the ``--generate_dirs`` option is used. The info can be test-specific
   or can be modified globally by wrapping the ``verify_image_cache`` fixture, e.g.:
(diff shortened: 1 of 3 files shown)


pytest_pyvista/pytest_pyvista.py
```diff
@@ -342,6 +343,27 @@ def _add_doc_cli_and_ini_options() -> None:
...
+    parser.addini(
+        "pyvista_blank_image_atol",
+        default=str(DEFAULT_BLANK_IMAGE_ATOL),
+        help=(
+            "Absolute tolerance for the blank image check. A rendered image whose "
+            f"difference from a blank reference is <= this value is flagged as blank "
+            f"(default: {DEFAULT_BLANK_IMAGE_ATOL})."
+        ),
+    )
+
```
user27182, line 367: - If you move this inside `_add_doc_cli_and_ini_options()` you also get CLI arg for free
- Can we not simplify this to a single parameter `pyvista_blank_image_atol`? Where setting it to a value at all enables it (and sets the tolerance value).
- I like the prefix, `pyvista_` but none of the other options use this (but maybe they should all be renamed?)
- I can see some legitimate docstrings having a blank image plot. Should we add a custom hook, similar to `pytest_pyvista_max_vtksz_file_size_hook`, to enable customizing this value on a per-test basis?

README.rst
```diff
@@ -344,6 +344,12 @@ in the beginning of your test function.
   ``verify_image_cache`` fixture is used but no images are generated. The value of this
   flag takes precedence over the global flag by the same name (see above).
 
+* ``allow_blank_image``: When the global ``pyvista_check_blank_images`` option is enabled
+  (see `Configuration`_), every test that renders an essentially blank (near-uniform)
+  image fails, since this usually indicates a broken test that renders nothing yet still
+  passes. Set this flag to ``True`` to bypass the check for a test that intentionally
+  renders a near-blank scene.
+
```
user27182, line 358: The ability to detect blank images seems like it should apply to unit tests _and_ doc mode tests. But it seems like this PR adds it for unit tests only. (the comment I made earlier https://github.com/pyvista/pytest-pyvista/pull/297/changes#r3256052545 I assumed it was _only_ for doc mode)

I suggest enabling it for both.

Related to https://github.com/pyvista/pytest-pyvista/pull/293#discussion_r3256168287

pytest_pyvista/pytest_pyvista.py
user27182: > * Can we not simplify this to a single parameter `pyvista_blank_image_atol`? Where setting it to a value at all enables it (and sets the tolerance value).

Or maybe the two parameters is fine after all for clarity. As long as in default case the `atol` is set such that, in general, users should not really need to use `pyvista_blank_image_atol`, that means most configs will set just one variable anyway: `pyvista_check_blank_images=True`.

PR #298, opened 2026-05-18
Add autouse fixtures to reset PyVista global state and theme

### LINKED ISSUE #277: Global State and Theme Reset
PyVista has global settings (`vtk_snake_case`, `vtk_verbosity`, `allow_new_attributes`, `PICKLE_FORMAT`) and a global theme that tests can modify. Without resetting between tests, one test's side effects leak into the next.

### What PyVista does

Two separate fixtures handle this in `tests/plotting/conftest.py` and `tests/conftest.py`:

```python
# Function-scoped: reset PyVista global state
@pytest.fixture(autouse=True)
def _reset_pyvista_state():
    yield
    pv.vtk_snake_case('error')
    pv.vtk_verbosity('info')
    pv.allow_new_attributes(False)
    pv.PICKLE_FORMAT = 'vtk'
```

```python
# Function-scoped: reset the plotting theme
@pytest.fixture(autouse=True)
def _set_default_theme():
    pv.global_theme.load_theme(pv.plotting.themes._TestingTheme())
    yield
    pv.global_theme.load_theme(pv.plotting.themes._TestingTheme())
```

The theme reset runs both before and after each test, ensuring a clean baseline for image regression and preventing one test's theme changes (background color, font size, edge visibility, etc.) from affecting the next.

### Proposal

Add both fixtures to pytest-pyvista. The state reset covers settings that any PyVista-based project might touch. The theme reset is especially important for projects using `verify_image_cache`, since a leaked theme change silently corrupts image baselines.

The state reset should be configurable since downstream projects may intentionally set different defaults:

```toml
[tool.pytest.ini_options]
pyvista_reset_global_state = true  # default: true
```

banesullivan:
PyVista global settings (`vtk_snake_case`, `vtk_verbosity`, `allow_new_attributes`, `PICKLE_FORMAT`) and the plotting theme leak between tests, so one test silently corrupts the next. Every PyVista-based suite reimplements the same reset fixtures. This adds them to the plugin: `_reset_pyvista_state` after each test, and `_set_default_theme` which loads the testing theme before and after a test but only when `verify_image_cache` is requested, so non-plotting suites stay fast (per @edabor's suggestion in the issue).

Both default on; the state reset is gated by `pyvista_reset_global_state`. Each reset is individually guarded so older PyVista degrades gracefully.

Blast radius: a project using `verify_image_cache` now renders under PyVista's `_TestingTheme`. Image baselines generated under a different theme may need regenerating. Doc-mode tests do not request `verify_image_cache`, so they keep the default theme and are unaffected.

resolves #277


pytest_pyvista/pytest_pyvista.py
```diff
@@ -341,6 +341,12 @@ def _add_doc_cli_and_ini_options() -> None:
         default=True,
         help="Automatically close all plotters and run gc.collect() after each test (default: True).",
     )
+    parser.addini(
+        "pyvista_reset_global_state",
+        type="bool",
+        default=True,
+        help="Reset PyVista global state (snake case, verbosity, attributes, pickle format) to defaults after each test (default: True).",
+    )
```
user27182: So this does not affect the plotting theme? Why are global config settings set with a CLI arg, but global plotting settings are set using the `verify_image_cache` fixture? Seems a bit inconsistent in terms of UI to set these similar global settings in very different ways.

I can foresee many of the global settings being moved into the new `global_config` upstream. And similarly a new  `_TestingConfig` may be added with the testing defaults. So perhaps we should use that language accordingly, and name this `pyvista_reset_global_config`.

user27182 (approved):

user27182:
I made some key changes:

* Split fixtures into `_reset_fixtures.py`/`test_reset_fixtures.py`
* Fixed bugs in the global-state reset defaults for pickle and `allow_new_attributes`
* Folded theme reset into `verify_image_cache` (was a separate fragile fixture)
* Dropped `pyvista_` prefix from the config name
* Follow existing conventions for adding new CLI/INI options

PR #304, opened 2026-08-12
Grade multi-baseline tests against the closest cached image

### LINKED ISSUE #286: multi-cache selection bug

### LINKED ISSUE #303: Multi-baseline rescue is skipped for warning-level mismatches, and never passes silently
### Summary

When a test has a directory of cached baseline images, the fallback comparison against the remaining images is gated on the *initial* comparison producing an **error** ([`pytest_pyvista.py#L542`](https://github.com/pyvista/pytest-pyvista/blob/main/pytest_pyvista/pytest_pyvista.py#L542)):

```python
if fail_msg and len(cached_image_paths) > 1:
```

That gate causes two surprising behaviours. In both, a later cached image is a **byte-exact match** for the render, and in both the test still emits a warning.

### Reproducer

```python
# make_cache.py
from pathlib import Path

import pyvista as pv

pv.OFF_SCREEN = True

d = Path("cache", "imcache")
d.mkdir(parents=True, exist_ok=True)
for color, name in (("red", "im1.png"), ("blue", "im2.png")):
    p = pv.Plotter()
    p.add_mesh(pv.Sphere(), color=color)
    p.screenshot(d / name)
```

```python
# test_repro.py
import pyvista as pv

pv.OFF_SCREEN = True


def test_imcache(verify_image_cache):
    # raise error_value so the mismatch against im1.png is a *warning*, not an error
    verify_image_cache.error_value = 100_000
    p = pv.Plotter()
    p.add_mesh(pv.Sphere(), color=[0, 0, 254])
    p.show()
```

```console
$ python make_cache.py
$ pytest test_repro.py --image_cache_dir cache
```

The render is an exact match for `im2.png` (`compare_images` → `0.0`), and mismatches `im1.png` by `62628.07`.

**Symptom 2 — a warning-level mismatch never consults the other images at all.**
As written above, the mismatch against `im1.png` exceeds `warning_value` but not `error_value`, so `fail_msg` is `None` and the loop is never entered:

```console
UserWarning: imcache Exceeded image regression warning of 200.0 with an image error of 62628.06535948396
```

The remaining baselines are never looked at, and the warning doesn't even mention that the test has multiple cached images.

**Symptom 1 — an exact match with a later image still warns.**
Delete the `error_value` line and rerun, so the first mismatch is an error. The rescue now runs and finds the exact match, but only ever *downgrades* the error to a warning:

```console
UserWarning: imcache Exceeded image regression error of 500.0 with an image error equal to: 62628.06535948396
This test has multiple cached images. It initially failed (as above) but passed when compared to:
	cache/imcache/im2.png
```

This one needs no configuration at all — it is the behaviour asserted today by `test_multiple_cache_images[ALMOST_BLUE]`.

### Expected

If the render is a clean match for *any* image in the cache directory, the test should pass silently. A directory of baselines means "any of these is acceptable"; a warning implies someone needs to act, and here nobody does.

### Relationship to #286 / #295

I think symptom 1 is the same thing reported in #286 ("triggering false-positive warnings even when a matching image exists"), seen from the other end. #295 fixes it by making the *right* image get compared first, which removes the warning whenever the correct ordering can be inferred from `_EnvInfo`. Removing the gate fixes it for the general case, including caches keyed on something `_EnvInfo` cannot model. The two are complementary and touch the same block, so whichever lands second will need a rebase.

### Why this matters downstream

We hit this in [geovista](https://github.com/bjlittle/geovista). PROJ 9.8 switched `+proj=eqc` to the ellipsoidal EPSG:1028 method (OSGeo/PROJ#4656), and conda-forge ships PROJ 9.8.1 while the PyPI `pyproj` wheels still bundle 9.5.1. We therefore keep two baselines per projected test, and they differ by:

```
compare_images("proj-lt-9.8.png", "proj-ge-9.8.png") = 251.18
```

Above `warning_value` (200), below `error_value` (500) — squarely in the gap. Because the rescue never runs, those tests warn on whichever leg isn't first alphabetically, and our suite runs with `filterwarnings = ["error"]`, so they fail outright.

#295 can't help us here: our variants are keyed on the PROJ version, which `_EnvInfo` (OS, GPU vendor, VTK version) has no notion of, so no ordering heuristic can put the right one first. We currently work around it with a `verify_image_cache` fixture override setting `error_value = warning_value`, purely to force `fail_msg` to exist so that the rescue loop runs at all.

### Suggested fix

Enter the loop for warnings as well as errors, and pass silently on a clean match:

```python
if (fail_msg or warn_msg) and len(cached_image_paths) > 1:
```

I have this working with tests — happy to open a PR, or to leave it for whoever picks up #295.

bjlittle:
Closes #286. Closes #303. Supersedes #295, which can be closed with this.

Implements the rule agreed in #303: *warn only if all of the images are above the
warning threshold, and error only if none of the images are under the error
threshold.*

### What

When a test has a directory of cached baselines, the comparison against the remaining
images was gated on the first comparison producing an error:

```python
if fail_msg and len(cached_image_paths) > 1:
```

Two consequences:

1. A mismatch landing between `warning_value` and `error_value` never consulted the
   other baselines at all — the warning was emitted without ever looking at them
   (#303).
2. An error that *was* rescued could only ever be demoted to a warning, even when the
   rescuing image was an exact match (#286).

This compares against every cached image and grades the test on the closest match. The
loop short circuits on the first image at or under `warning_value`, so a correctly
cached test still costs exactly one comparison — only a test that is already going to
warn or fail pays for the rest.

`doc_mode.test_images` carried a second copy of the same loop, so the whole comparison
now lives in `_test_compare_images` and both call sites share it. That also drops a
small inconsistency in doc mode, where the first comparison went through
`_compare_images` but the retries called `pv.compare_images(pv.read(...), ...)`
directly.

### Behaviour

| closest match | before | after |
|---|---|---|
| under `warning_value` | pass silently *(only if it was the first image)* | pass silently |
| between the thresholds | warning | warning |
| over `error_value` | fail | fail |

The grade now depends only on the *set* of cached images, not on which one happens to
sort first. `_get_file_paths` returns them sorted, so previously a cache whose ordering
didn't match the environment warned on every healthy run — the motivating case in #303
is a `+proj=eqc` baseline pair that differ from each other by 251.18, against the
default thresholds of 200/500.

Single-image caches are unaffected: the loop runs once and grades on that image, as
before.

One knock-on: `--reset_only_failed` now resets the closest-matching cached image rather
than always the first one in the directory, which seems like the more useful variant to
update. Single-image caches are unchanged.

### Messages

`--failed_image_dir` and the warning/error text now report the closest match rather
than the first image compared:

```
imcache Exceeded image regression error of 500.0 with an image error equal to: 806.7
This test has multiple cached images. The closest match was:
	cache/imcache/im2.png
```

replacing `It initially failed (as above) but passed when compared to:` /
`and failed again for all images in:`, neither of which describes what now happens.

The paragraph in `README.rst` describing the old first-then-the-rest algorithm has been
rewritten to match.

### Tests

Mirrored in `test_pyvista.py` and `test_doc_mode.py`:

* New `test_multiple_cache_images_clean_match`, parametrised over an initial mismatch
  that errors and one that only warns. Both must pass silently. All four
  parametrisations fail on `main`.
* `test_multiple_cache_images` keeps its structure and its `--failed_image_dir`
  coverage. Its `ALMOST_BLUE` case exercised the error-demoted-to-warning path by
  relying on the second cached image being an exact match — which now passes silently —
  so the warning threshold is lowered to `-1.0`, making that image the closest match
  but not a clean one, which is the warning case under the new rule. `ALMOST_RED`
  additionally asserts that a clean match against the first image says nothing at all.

Verified against the full suite: the set of non-passing tests is byte-identical to
`main`, plus the four new passes. (My machine has no `playwright` browsers installed,
so `check_playwright_config` was stubbed out identically in both runs; the `vtksz`
tests fail either way and are unaffected by this change.)

### On staleness

Per the discussion in #303 this doesn't try to detect stale cache entries.
`--disallow_unused_cache` is the natural home for that and can't do it today —
`cached_image_names` is built from a non-recursive `cache_path.glob(f"*.{image_format}")`
so variant subdirectories are never candidates, and `VISITED_CACHED_IMAGE_NAMES` records
the test's image name rather than the variant that matched. Happy to look at that
separately if it's wanted.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -204,7 +204,7 @@ jobs:
 
   downstream:
     name: Downstream tests
-    runs-on: ubuntu-22.04-self-hosted # matching pyvista
+    runs-on: ubuntu-22.04 # matching pyvista
     steps:
       - uses: actions/checkout@v6
 
--- a/README.rst
+++ b/README.rst
@@ -94,11 +94,17 @@ Can be replaced with multiple versions of the image:
    - Use the ``--generate_subdirs`` flag to automatically generate test images in a
      sub-directory format.
 
-When there are multiple images, the test will initially compare the test image
-to the first cached image. If that comparison fails, the test image is then
-compared to all other cached images for that test. The test is successful if one
-of the comparisons is successful, though a warning is still issued if it initially
-failed.
+When there are multiple images, the test image is compared to each of the cached
+images and is graded against the closest match. A warning is only issued if every
+cached image exceeds the warning threshold, and the test only fails if every
+cached image also exceeds the error threshold. A test which closely matches any
+one of its cached images therefore passes silently, regardless of the order in
+which the images are compared.
+
+.. note::
+
+   The comparisons stop as soon as an image is matched within the warning
+   threshold, so a passing test is typically only compared to a single image.
 
 Both use cases (i.e. unit tests and documentation tests) support specifying multiple
 cache images.
--- a/pytest_pyvista/doc_mode.py
+++ b/pytest_pyvista/doc_mode.py
@@ -24,7 +24,6 @@ from .pytest_pyvista import DEFAULT_WARNING_THRESHOLD
 from .pytest_pyvista import PYVISTA_FAILED_IMAGE_CACHE_DIRNAME
 from .pytest_pyvista import PYVISTA_GENERATED_IMAGE_CACHE_DIRNAME
 from .pytest_pyvista import _AllowedImageFormats
-from .pytest_pyvista import _check_compare_fail
 from .pytest_pyvista import _ensure_dir_exists
 from .pytest_pyvista import _EnvInfo
 from .pytest_pyvista import _get_file_paths
@@ -534,7 +533,6 @@ def test_images(_pytest_pyvista_test_case: _DocVerifyImageCache, doc_verify_imag
     cached_input_is_file = cached_image_path.is_file()
     cached_image_paths = [cached_image_path] if cached_input_is_file else _get_file_paths(cached_image_path, ext=_DocVerifyImageCache.image_format)
     cached_image_paths = cast("list[Path]", cached_image_paths)
-    current_cached_image_path = cached_image_paths[0]
 
     # Ensure test path is an image
     test_image_path = cast("Path", test_case.test_image_path)
@@ -546,28 +544,13 @@ def test_images(_pytest_pyvista_test_case: _DocVerifyImageCache, doc_verify_imag
             test_image_path.rename(new_path)
             test_image_path = new_path
 
-    warn_msg, fail_msg = _test_compare_images(
+    warn_msg, fail_msg, current_cached_image_path = _test_compare_images(
         test_name=test_case.test_name,
         test_image=test_image_path,
-        cached_image=current_cached_image_path,
+        cached_image_paths=cached_image_paths,
         allowed_error=DEFAULT_ERROR_THRESHOLD,
         allowed_warning=DEFAULT_WARNING_THRESHOLD,
     )
-
-    # Try again and compare with other cached images
-    if fail_msg and len(cached_image_paths) > 1:
-        # Compare build image to other known valid versions
-        msg_start = "This test has multiple cached images. It initially failed (as above)"
-        for path in cached_image_paths[1:]:
-            error = pv.compare_images(pv.read(test_image_path), pv.read(path))
-            if _check_compare_fail(test_case.test_name, error, allowed_error=DEFAULT_ERROR_THRESHOLD) is None:
-                # Convert failure into a warning
-                warn_msg = fail_msg + (f"\n{msg_start} but passed when compared to:\n\t{path}")
-                fail_msg = None
-                current_cached_image_path = path
-                break
-        else:  # Loop completed - test still fails
-            fail_msg += f"\n{msg_start} and failed again for all images in:\n\t{_DocVerifyImageCache.image_cache_dir / test_case.test_name!s}"
 
     if fail_msg:
         _save_failed_test_image(test_image_path, "errors")
--- a/pytest_pyvista/pytest_pyvista.py
+++ b/pytest_pyvista/pytest_pyvista.py
@@ -454,7 +454,7 @@ class VerifyImageCache:
         skip_macos = platform.system() == "Darwin" and macos_skip_image_cache
         return skip or ignore_image_cache or skip_windows or skip_macos
 
-    def __call__(self, plotter: Plotter) -> None:  # noqa: C901, PLR0912, PLR0915
+    def __call__(self, plotter: Plotter) -> None:  # noqa: C901, PLR0912
         """
         Either store or validate an image.
 
@@ -530,28 +530,13 @@ class VerifyImageCache:
             return
 
         test_name_no_prefix = test_name.removeprefix("test_")
-        warn_msg, fail_msg = _test_compare_images(
+        warn_msg, fail_msg, current_cached_image = _test_compare_images(
             test_name=test_name_no_prefix,
             test_image=plotter,
-            cached_image=current_cached_image,
+            cached_image_paths=cached_image_paths,
             allowed_error=allowed_error,
             allowed_warning=allowed_warning,
         )
-
-        # Try again and compare with other cached images
-        if fail_msg and len(cached_image_paths) > 1:
-            # Compare test image to other known valid versions
-            msg_start = "This test has multiple cached images. It initially failed (as above)"
-            for path in cached_image_paths[1:]:
-                error = _compare_images(plotter, path)
-                if _check_compare_fail(test_name, error, allowed_error=allowed_error) is None:
-                    # Convert failure into a warning
-                    warn_msg = fail_msg + (f"\n{msg_start} but passed when compared to:\n\t{path}")
-                    fail_msg = None
-                    current_cached_image = path
-                    break
-            else:  # Loop completed - test still fails
-                fail_msg += f"\n{msg_start} and failed again for all images in:\n\t{Path(self.cache_dir, test_name_no_prefix)!s}"
 
         if fail_msg:
             if self.failed_image_dir is not None:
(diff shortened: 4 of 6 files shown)


user27182 (approved):

user27182:
Will need to fix the self-hosted runner issue first (move back to gh hosted)

PR #305, opened 2026-08-19
Keep running trame servers in the registry

larsoner:
Clearing the whole registry after each test does not stop a running server: trame starts another one on the next request, while the first keeps its asyncio task and protocol graph alive, and trame_vtk's HELPERS_PER_SERVER -- keyed by server name, not by server -- strands the old helper with its vtkWebApplication. Measured over three exporting tests, one vtkWebApplication and three Server objects leaked per test; with running servers kept, the helper and its application are reused and the counts stay flat.

Idle servers are still dropped, which is what prevents the collision the registry clearing was added for in #278.

Surfaced by pyvista/pyvista#8929, and by the leak check blocking #304. cc @user27182 

One decision point:

- Should a running server be kept, or actually stopped? My patch keeps it, so one live server and its serializer caches persist for the whole session — less isolation than the current code intends. But the current code doesn't achieve isolation either: it only forgets the server, which keeps running, so the next use starts a second one and strands the first (measured: one vtkWebApplication and three Server objects leaked per exporting test). The isolation-preserving alternative is to stop the servers rather than forget them — but trame_server.core.Server.stop() is a coroutine, so a synchronous teardown fixture would have to drive trame's event loop, which is more machinery and can hang.

Changes drafted by Claude Opus 5 and reviewed/understood by me.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -204,7 +204,7 @@ jobs:
 
   downstream:
     name: Downstream tests
-    runs-on: ubuntu-22.04-self-hosted # matching pyvista
+    runs-on: ubuntu-22.04 # matching pyvista
     steps:
       - uses: actions/checkout@v6
 
--- a/pytest_pyvista/pytest_pyvista.py
+++ b/pytest_pyvista/pytest_pyvista.py
@@ -1088,7 +1088,7 @@ def _close_plotters_clear_trame_servers(pytestconfig: pytest.Config) -> Generato
 
     This teardown fixture serves mutltiple purposes:
     - closing all plotters,
-    - clearing trame servers registry (to prevent test collision),
+    - clearing idle trame servers from the registry (to prevent test collision),
     - forcing garbage collection.
     """
     yield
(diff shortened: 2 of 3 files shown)


user27182:
Nice! Just curious, does this duplicate any test fixture or cleanup in `pyvista` itself? As in, when `pytest-pyvista` gets bumped downstream by `pyvista`, is there anything there we can cleanup and remove?

Also you'll need to drop self-hosted, like https://github.com/pyvista/pytest-pyvista/pull/304/commits/68d2ea30d645bc72d56faf8e1c28c184c9aba785

user27182:
> Should a running server be kept, or actually stopped?

I'm not sure. Maybe status-quo unless/until another issue pops up?

larsoner:
I asked Claude and its analysis seems reasonable:

---

Deletable on bump: `macos_memory_leak` in pyvista's `tests/plotting/conftest.py:56` is the same fixture as `_pyvista_macos_autorelease` here, down to the NSAutoreleasePool dance and the VTK issue it cites.

I'd keep: the repo-wide `pv.close_all()` in pyvista's `tests/conftest.py:396`. It overlaps with `pyvista_close_all` here, and the ordering does work, but pyvista's leak check depends on plotters being closed before it scans, and an ini option in a third-party plugin shouldn't be able to silently turn that off.

Not duplicated: the trame fixtures in `tests/plotting/jupyter/conftest.py` (`_eager_resolve_trame_component` warms the server before any leak snapshot, `_trame_array_cache` clears trame's 20-second serializer cache). They're complementary — though this PR does make the warm-up actually stick, since the server stays registered.

Worth noting pyvista pins `pytest-pyvista==0.3.3`, which predates #278 entirely, so none of this is live downstream yet.

larsoner:
Okay this needs a PyVista-side change, will open a PR there...

jourdain:
FYI, you can request to stop a trame server if that helps. It will need to be async but that should stop it cleanly.

larsoner:
Yes the stop is a good first step, but trame-vtk keeps a reference to the vtk object in `CoreServer.sharedObjects["app"]`, so the VTK objects appear as "leaks" to the test infra. I'll see if a manual clear fixes things....

larsoner:
Okay I think this should be good to go. Perhaps fragile since we're messing with private attrs but it allows the objects to be GC'ed.

user27182 (approved):

user27182:
Probably needs a patch release too and bump on pyvista side to resolve https://github.com/pyvista/pyvista/issues/8929 ?

larsoner:
I think 8929 must be separate because it still pins 0.3.3 and this fixes stuff only in `main` here

larsoner:
... so I think this is needed before cutting a release, and something separate is needed for 8929 (working on it still...)

user27182:
Thanks for confirming! Doesn’t look like this is holding up anything upstream, so will simply merge, no need for a release yet

PR #306, opened 2026-08-20
Speed up test suite: drop matplotlib import, shrink xdist parallel-image test

user27182:
Minor optimizations to reduce test times:
Baseline: 304 tests, ~133s.
This branch: ~121–126s locally

- Hardcode the matplotlib tab10 color cycle instead of importing matplotlib just to read it, removing an unused heavy dependency import.
- Render images at 100x100 instead of the pyvista default 1024x768 in test_multiple_cache_images_parallel (non-vtksz variant), since the test only compares images to each other, not their fidelity.


user27182:
Only seeing a few seconds difference relative to main.

PR #309, opened 2026-08-24
Add an interactive HTML image summary report with per-image baseline approval

bjlittle:
I've been meaning to create something like this for a long while.

This is a proof-of-concept and it's completely created using `claude`, so it's cheap for me to tweak it how we want ... so consider it a strawman, but I think it adds value and fills an obvious gap in our workflow.  

See the [design](https://github.com/pyvista/pytest-pyvista/pull/309/changes#diff-163c354797e656a9a147f30a8f476c8a7c0790d3402d0858d250315b3c037c94) for fuller context. Note that I've included the `doc/superpowers` contents as part of this PR for clarification - we don't necessarily need to bank it; something to be decided.

## Problem

When an image test fails, `pytest-pyvista` tells you an error value exceeded a threshold. It doesn't show you the images. Deciding whether a regression is a genuine break or an acceptable rendering drift means finding the baseline, finding the generated render, and comparing them by hand — and by default nothing is even persisted to compare, because the generated images live in the pytest cache and are deleted at the end of the run.

Updating a baseline is equally coarse: `--reset_image_cache` overwrites everything, `--reset_only_failed` overwrites every failure. There is no way to say "this one drifted acceptably, take it; that one is a real bug, leave it".

## What this adds

An opt-in HTML report, generated at the end of a normal run, that catalogues **every** image test — not just the failures — as a card showing the cached baseline, the generated render, and the pixel difference, alongside the error value and the threshold it was actually measured against. From the report you can approve individual images and apply them to the cache with a new console script.

```bash
pytest --summary_html
# -> image_test_report/index.html
```

### Benefits

- **See the regression instead of reading a number.** A magenta pixel-difference overlay on a faded greyscale baseline shows exactly what moved.
- **Per-image baseline updates.** Approve the drift, reject the bug, in one pass — instead of an all-or-nothing cache reset.
- **The whole suite, not just failures.** Passing images are catalogued too, so drift can be spotted before it crosses a threshold.
- **CI-friendly.** The report is a self-contained directory (or a single file with `--summary_html_embed`) that can be uploaded as a build artifact and opened later.
- **Zero cost when off, and no new dependencies.** Disabled by default; PIL and NumPy already arrive via `pyvista`.

## How it works

Records are captured **inside** `VerifyImageCache.__call__` at comparison time, not reconstructed by scanning directories afterwards. Three facts force this: nothing is persisted by default; the cache-writing policies destroy the prior baseline before it can be read; and only the comparison itself knows which threshold was in force for that specific call.

Each xdist worker appends `ImageRecord` JSONL to a temp directory. The master combines them in `pytest_terminal_summary` — before `pytest_unconfigure` wipes the records — and renders the report. The baseline is copied aside *before* any cache write, so under `--reset_image_cache` the report still diffs against the true prior baseline rather than the image against itself.

Report generation can never change a test outcome: every capture point and the terminal-summary write are guarded, and a failure degrades to a warning naming the cause, leaving the exit code and terminal summary untouched.

### Layout

```
pytest_pyvista/summary/
├── record.py    ImageRecord, statuses, JSONL write/read/combine
├── diff.py      pixel-difference image, size-mismatch detection
├── store.py     report image store: slugging, downscaling, full-size retention
├── collect.py   status determination (pure logic, no pytest session needed)
├── session.py   per-session capture, called from VerifyImageCache
├── render.py    records -> index.html
├── approve.py   manifest validation + the pytest-pyvista-approve CLI
└── assets/      report.css, report.js (vanilla, no frameworks, no CDN)
```

## What the report includes

Every generated image becomes a card carrying one of six statuses:

| Status | Meaning |
|---|---|
| `passed` | The render matches its baseline. |
| `warned` | Error exceeded the warning threshold but not the error threshold — the test passed, but the image is drifting. |
| `failed` | Error exceeded the error threshold. |
| `skipped` | Comparison was skipped, so no render was generated. |
| `new` | No baseline existed for this image. |
| `reset` | A baseline existed and was overwritten this run by `--reset_image_cache` / `--reset_only_failed`. The report compares against the baseline as it stood *before* the overwrite. |

Each card shows the baseline, the generated render and the pixel difference, the error against the threshold actually in force (including per-test `var_error_value` / `var_warning_value` overrides and `high_variance_test`), which of several candidate baselines matched, and — where relevant — that the image was written to the cache during the run and why. Size mismatches are reported with both dimensions rather than a broken diff.

The page itself provides status filters, a test-name search, per-image approve checkboxes, **Accept all new**, and **Export approvals**. It is vanilla CSS and JS, theme-aware, and renders offline.

### Flags

| Flag | Purpose |
|---|---|
| `--summary_html` | Enable report generation. |
| `--summary_html_dir DIR` | Report directory (default `image_test_report`). Implies `--summary_html`. |
| `--summary_html_include STATUSES` | Comma-separated statuses to include. |
| `--summary_html_max_image_size N` | Longest-edge pixel limit for inline images (default 400). |
| `--summary_html_full_size {none,failing,all}` | Which records retain a full-resolution copy (default `failing`). |
| `--summary_html_embed` | Embed images as data URIs, producing a single HTML file. |

All have matching ini options.

## Developer workflow

```bash
# 1. Run the suite and get a report
pytest --summary_html

# 2. Open image_test_report/index.html, review the cards,
#    tick the images whose new render should become the baseline,
#    then click "Export approvals" -> approvals.json

# 3. Preview what would happen
pytest-pyvista-approve approvals.json --target cache --dry-run

# 4. Apply
pytest-pyvista-approve approvals.json --target cache

# 5. Re-run: the approved images now pass
pytest
```

`--target staging` (the default) copies into a staging directory for a second look instead of writing to the cache. `--image_cache_dir` names the project's real cache when it isn't at the default; `--generated_image_dir` does the same for the sources.

The manifest is treated as untrusted input, since it round-trips through a browser's download directory: unknown schema versions are rejected, the declared cache directory is checked against the real one, and every source and destination must resolve inside its permitted root — no traversal, no absolute paths outside those roots, no symlinks followed out. Copies are atomic (temp file plus `os.replace`), so a failed copy can never destroy an existing baseline.

## Out of scope

- **`--doc_mode`.** The summary flags are rejected under doc mode; the two features are not integrated.
- **A JavaScript test runner.** The report's client-side behaviour is covered by string-containment smoke checks in the committed suite. Adding Playwright was considered and declined; the JS was verified during development with a throwaway Node harness driving the real shipped `report.js` against real rendered pages, which is deliberately not committed.
- **Optimising the capture path.** Measured at ~79 ms per image, roughly 1.3 minutes added to a 1000-image suite, with a report directory between ~36 MB (mostly passing) and ~112 MB (all failing). Judged acceptable at PyVista's scale; left alone rather than optimised speculatively.
- **Cleaning stale renders.** `image_test_report/generated/` is not pruned between runs, matching existing `images/` behaviour. Documented.

## Follow-ups

- **`reset` status coverage.** Five of the six statuses are covered by end-to-end integration tests; `reset` is not, because `--reset_image_cache` applies to a whole run and can't be mixed into a single-run fixture.
- **Edge-diff legibility.** A wireframe change covers ~0.6% of pixels at the default 400px thumbnail versus ~4.6% at full resolution. Worth deciding whether the diff overlay should dilate changed pixels so fine-detail regressions are visible while scanning.
- **Manifest contract single-sourcing.** The payload shape is duplicated across `ImageRecord`, `render.py`'s `_manifest()`, `report.js`'s `exportApprovals()` and `approve.py`'s `_REQUIRED_KEYS`; only `SCHEMA_VERSION` is genuinely shared. A round-trip test guards drift, but the constant itself could be single-sourced.
- **`SCHEMA_VERSION` split.** One constant currently versions both the JSONL record format and the export manifest. A records-only change would either invalidate previously exported manifests or tempt skipping the bump. Both are `1` and nothing is deployed, so this is cheap to do later.
- **Minor validation hardening in `approve.py`:** `test_name` / `image_name` are not type-checked (neither builds a path today); a manifest omitting `approved` loads as empty rather than being rejected; `run_id`, `exported_at` and `image_format` are carried but never validated.
- **Flag naming.** `--staging_dir` / `--generated_image_dir` / `--image_cache_dir` are snake_case while `--dry-run` is kebab-case, following the plan as written. Worth unifying before the CLI has users.

## Testing

- 195 new tests across 11 modules, plus the existing suite: **375 passed, 1 skipped** (the single pre-existing `test_clear_trame_servers` failure is unrelated — it needs the trame extra).
- Integration tests drive real `pytester` sessions with real renders, covering xdist single-report generation, the `--disallow_unused_cache` abort interaction, embed mode, full-size retention modes, subdirectory baselines, and the render → export → apply round trip.
- The applier's path containment was probed with 60+ hand-built hostile manifests across review rounds: `..` traversal, symlink escapes in both directions, symlinked destination parents, string-prefix sibling directories, symlink loops, destination-equals-root, and embedded NUL bytes.
- `ruff` clean at `select = ["ALL"]`, `ruff format --check` clean, and `mypy` unchanged at its pre-existing baseline with no errors in the new modules. No lint or type suppressions were added anywhere in the branch.

## Demo

https://github.com/user-attachments/assets/e50ebaa4-d6f9-447f-8952-994d663b248c

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)


user27182:
Thanks @bjlittle !

I like the idea, we already do something like this in PyVista, where each PR has test reports uploaded:
https://6a8ddb5135f8ead89b480f69--pyvista-doctest-images.netlify.app
https://6a8ddab6b461ea27abc38de8--pyvista-test-images.netlify.app

They show up in the checks for every PR:
image

Not sure if you had a look at _that_ report, but maybe we can merge features from that html report into your new report?
Basically, I'd like it if we could simply re-use the new reports without needing to make too many changes downstream. It looks like your reports already have lots of the features (incl. filtering out warnings), but I think some key things might be missing? Like support for `--doc_mode`, as well as support for multiple reports from multiple runners across OS versions and vtk versions.

Also, this PR is quite massive (and so is the description). Would it be possible to split this out into smaller PRs?
(1) core machinery needed to upload a super basic report with warnings and errors
(2) extra bells and whistles, html-only features like filters and difference image

Or maybe even easier is to just ask Claude to automatically split it up for you without dictating exactly how. GitHub has a new stacked PRs feature, maybe you can have it open up several new stacked PRs accordingly?

bjlittle:
@user27182 I'll close and re-think and re-spin.

user27182:
Thanks! I do really like this and want to see it move forward. Just a few too many features in one PR I think for me to give it a proper review.

bjlittle:
@user27182 Thanks. I'll craft some time to iterate on this, don't worry I've not forgotten about this and don't want it to fall between the cracks :100:

PR #310, opened 2026-08-31
Report an invalid image cache once per run, not once per test

user27182:
`InvalidCacheError` came from a session-scoped fixture, and pytest re-raises a failed fixture's exception for every test requesting it — so one malformed cache errored every test in the suite.

Now validated once in `pytest_configure` and reported from `pytest_collection` as a single collection error that stops the run early. Message and exception type unchanged; exit code becomes 2 (`INTERRUPTED`). `pytest` floor bumped to 7.0.

🤖 [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -17,7 +17,7 @@ classifiers = [
 ]
 dependencies = [
   "pyobjc-framework-Cocoa; sys_platform == 'darwin'",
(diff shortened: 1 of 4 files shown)


user27182:
Superseded by #311 — the branch rename closed this one.

